PR #19, opened 2021-07-08
Added draft of bleedthrough

### LINKED ISSUE #12: Create New Bleed-Through Augmentation
**Bleed-through** is caused by the seeping of ink from the reverse side of the page, through the sheet of paper, and into the front side of the page.  This faint bleed-through effect can be seen from a duplex-printed document, ink pressed document, or simply a document scanned or copied from a thin paper source.

## Examples
image
image
image

## Reconstructing Bleed-Through Example
Using these sample images, we can construct the intuition for how we can simulate the bleed-through effect:

1. **Fabricate a Back Side Image** -- Use the source image (Front Side) to create a copy of the same image that is flipped to represent the reverse side of the page (Back Side).  Since we don't have any other source of information to draw from, we are forced to derive a plausible reverse side by taking the front side and reversing it.  We should do some random vertical shift of the content so that the horizontal content is not exactly the same as the front side.
2. **Degrade Back Side Image** -- Apply something like an aggressive Ink Bleed effect to the Back Side image.  Then, erode the ink as well as other treatments to reduce the intensity of the ink.
3. **Merge the Front Side + Back Side** -- Combine the layers bank into a single ink layer to simulate the ink-bleed effect.

**FRONT**
image
**BACK**
image
**MERGED FRONT + BACK**
image

kwcckw:
This is a draft of bleedthrough augmentation using combination of inkbleed and gaussian blur augmentations. Some sections need to be further discussed and improvised. 

For example:

1. Flipping the original image is not working for rotated inputs, so no flipping in current implementation until there's a better workaround.
2. Final image is computed from the mean of blurred and input images, it might be appeared darken a bit. Suggestions are welcome to improve on this issue. 

Example outputs with just bleed through augmentation:

Example1
image

Example2
image


jboarman:
@kwcckw This probably covers the basic implementation of the desired feature.  I don't understand why rotation causes any issue with the flipping, unless we are talking about images rotated in an augmentation earlier in the pipeline.  I think it would make sense to add a `flip_reverse_side` parameter that is defaulted to `True`.  Other than this optional flipping parameter, I think we can merge this in and improve upon it in future PRs.

In a future version of this augmentation, I think we should make the bleed through levels vary a bit more to model the concept of ink seeping through the page at different strengths based on varying wetness, paper thickness, texture, etc.  One approach for that would be to create an alpha mask using a [random noise generator](https://github.com/tatarize/noisemaker) and then subtracting ink from that reverse side ink layer using that alpha mask.

At some point soon-ish, we are going to have to deal with the fact that various augmentations involve a blur effect.  That effect can generally only be applied once per image, which is why the default pipeline only includes it once at the end.  We might potentially solve this with having more than one ink layer (at least for this particular augmentation) as well as tracking blur state within the pipeline payload data for a delayed invocation of the blur augmentation whereby augmentations can trigger that blur action to be queued up and fired at the end.  But again, that blur will only be useful against certain ink layers, and currently we only have a single ink layer.

@proofconstruction I think you may have some comments as it pertains to the recent shift towards some newer coding standards.  Since we are still early in devising and codifying those standards, I suggest we merge this in and continue to improve this code through additional PRs.

jboarman:
@kwcckw Using our new/current taxonomy (still likely a moving target), I believe this should be moved down into the `physical` folder in the just-updated version of the code (see #18).  Can you pull down the upstream changes that were just merged and move that augmentation into that folder?

kwcckw:
> 
> 
> @kwcckw Using our new/current taxonomy (still likely a moving target), I believe this should be moved down into the `physical` folder in the just-updated version of the code (see #18). Can you pull down the upstream changes that were just merged and move that augmentation into that folder?

Sure, will do that later.



> I don't understand why rotation causes any issue with the flipping, unless we are talking about images rotated in an augmentation earlier in the pipeline.

Right, in the current pipeline, images are rotated by default randomly:
image

Here the example of rotated output if we flip those rotated image directly:
image



> In a future version of this augmentation, I think we should make the bleed through levels vary a bit more to model the concept of ink seeping through the page at different strengths based on varying wetness, paper thickness, texture, etc. One approach for that would be to create an alpha mask using a random noise generator and then subtracting ink from that reverse side ink layer using that alpha mask.

Thanks, this should be a good idea to tackle on the current problem where i just sum both of the images.

jboarman:
Regarding the rotational effect on the flip action, I think it is fair to require that this augmentation be performed in the pipeline prior to any rotation.

kwcckw:
> 
> 
> Regarding the rotational effect on the flip action, I think it is fair to require that this augmentation be performed in the pipeline prior to any rotation.

Yes, that's one of the workaround and we would need to set the rotation as optional, since not all augmentations require rotation.

kwcckw:
I guess i will continue the development from my end here until we finalized #22 and i will create the pull request again.

proofconstruction:
To get a fully faithful representation of bleedthrough, we should be able to capture the rotation angle during the augmentation, then apply that rotation in the negative direction to the “back” image. If the rotation wasn’t performed, the bleedthrough effect should appear similar to the first picture above (only, without the rotation), but with the text reversed. If the text layer is rotated, it should be possible for the user to also rotate the reverse text separately, simulating a bad double-sided print.

I agree that the degree of bleedthrough in different parts of the image should be variable, simulating situations where too-much ink was applied (like the first time using a fountain pen). The first few images above, where the orientation of the text in both layers is the same, reminds me a lot of double-sided vellum prints, or books with [extremely thin “onion” paper](https://i.stack.imgur.com/EaduE.jpg). I think it’s pretty striking and may even deserve its own augmentation in the end.

The code itself is in line with what we currently have, but the last block needs some more documentation explaining what the shift functions are doing, and the long lines should be broken up. In general we need to be better about explaining large sections of numpy or cv2 calls, and making sure lines aren’t longer than 100 characters or so. This improves readability and makes it easier for other people to participate in the project either as users or contributors.

kwcckw:
Thanks for the suggestion, I will change it accordingly in the next update. 

Regarding the shift function, it is use to shift the ink so that it wouldn't be inline with those sentence. I will use a different approach in the next update so it should be more readable after this.

jboarman:
One minor improvement to diminish the ink on the back side might be simply to run it through `cv2.dilate` before blurring it.  Even if we also add some noise function, this might give it a slightly more natural appearance.

tbrettallen:
The issue you presented should be solvable with a horizontal+vertical flip instead of a rotation. I've attached an image that demonstrates this with a couple examples.
Flipping Solution

jboarman:
That’s brilliant!  Instead of a single flip, just flip both ways.  And that seems to work for just about any rotation angle.

kwcckw:
> 
> 
> The issue you presented should be solvable with a horizontal+vertical flip instead of a rotation. I've attached an image that demonstrates this with a couple examples.


Thanks, this is working for any rotation too. Just one thing to take note is right now those words are upside down, which may not occur in normal paper, but i guess this is the best workaround for now.

image

kwcckw:
> 
> 
> One minor improvement to diminish the ink on the back side might be simply to run it through `cv2.dilate` before blurring it. Even if we also add some noise function, this might give it a slightly more natural appearance.

Yes, and there's salt and pepper noise effect in the code and I think it would be more natural if we apply lowinkblob effect too, so the workflow would be:

dilate->add noise->lowinkblob ->gaussian

Generated bleedthrough ink (Before blending and gaussian):
image

After blending and gaussian blur:
image

What do you think on this?

jboarman:
That looks great.  I'll go ahead and merge in the PR (#24), and @proofconstruction can address an conflicts with the separate composability refactoring that he is working on.

kwcckw:
> 
> 
> That looks great. I'll go ahead and merge in the PR (#24), and @proofconstruction can address an conflicts with the separate composability refactoring that he is working on.

Alright, and i will add another update later which includes dilation and lowinkblob effect. Right now PR(#24) doesn't having these 2 effects yet.

PR #24, opened 2021-07-10
Updated bleedthrough augmentation.

### LINKED ISSUE #12: Create New Bleed-Through Augmentation

kwcckw:


proofconstruction (commented):
This looks good to me, and it’s a great example of the kind of code style I think we should move towards: separate functions for important operations, and a series of calls of those functions in the` __call__` method.

Issue #9, opened 2021-07-06
Standardize implementation of Randomization to follow Albumentations


kwcckw:
### Updated Issue Description

To improve composability of pipelines and to follow current standard practices exhibited by Albumentations, we should document and update our randomness functionality:

* [x] Document support for Python's [random seed](https://github.com/albumentations-team/albumentations/issues/93) for reproducible results within transforms and pipeline composition
* [x] Document "how to" to ensure all augmentation code shares Python's random state, even if Numpy random methods are needed in internally (see [example](https://github.com/sparkfish/augraphy/blob/9e4e11ce023a6ff0cbb71003869e490cffd068be/src/augraphy/base/oneof.py#L19))
* [x] Review all library's committed code for consistency with this standard
* [x] Change to use of `p` as probability parameter for consistency with Albumentations

See Pytorch notes on [reproducibility](https://pytorch.org/docs/stable/notes/randomness.html) for ideas on what we need to document.

#### Original Issue

> I tried to generate consistent output from the image but looks like it doesn't seem to be so straight forward. So wouldn't it be better if we making those randomly initialized parameters as optional? So that we can get consistent output with the same pipeline. 
> 
> Right now i can see the problem is there are so many randomize parameters and it would be difficult to get consistent output style from a same pipeline, so if i want to use the same pipeline with multiple papers, i might end up with different inconsistent style output.

shaheryar1:
Actually there is a way to remove randomization in the current code. To must apply an Augmentation you can set the `probability=1` or use` force=True`. Also most of the Augmentation use a `range=(lower_limit,higher_limit)` to select a value that controls the augmentation intensity. You simply can narrow down the range to get consistent results, or may be keep `lower_limit=upper_limit` .

jboarman:
@kwcckw Are you OK with the suggestion from @shaheryar1 on minimizing the randomness on any given augmentation?  I could see some other approaches used, including the introduction of a fixed seed value, which might make for more consistent testing and reproducible results.  What do you think?

kwcckw:
Yes, adjusting seed, probability,or use fixed range value would helps, but it wouldn't be straightforward since we need to adjust them one by one. I'm suggesting a general input argument to let user fixed the style or randomness, so that they can reproduce the same consistent output when needed.

jboarman:
[...]

Issue #22, opened 2021-07-09
The Road to Composability


proofconstruction:
Here are some thoughts about the architecture of the project going forward, from the perspective of composability. In the next 4 paragraphs, I propose some changes to the design of the base classes.

**Function-level composability**
Few of the existing Augmentation modules use more than the `__call__` method to perform their transformations. All of this behavior should be factored out into other functions that can then be called in the `__call__` method in sequence, or as the composition of those functions. These functions ideally would all be written in the "pure" style, where they operate only on their inputs and return a result, rather than operating on state defined elsewhere. This will make debugging easier and will also most likely surface some functions as candidates for refactoring into another library. As an example, the `create_scanline_mask` method in [DirtyRollersAugmentation](https://github.com/sparkfish/augraphy/blob/dev/src/augraphy/augmentations/physical/dirtyrollers.py#L91) is a prime candidate for splitting up into smaller functions.

**Module-level composability**
The current modules are all written in a tightly coupled style, where operations are applied to input images and then stored directly in a pipeline phase, from the `__call__` method of the augmentation. This should be changed so the call method instead takes an image as argument, then returns itself and the result of the completed operation as a tuple or list. Then the calling pipeline can sequence these on its own, without the augmentations being able to "see" the pipeline they're a part of. We would then be free to modify the internal behavior of augmentations or pipelines, so long as we maintain their interfaces.

This would facilitate direct composition of Augmentations too; in this case, it would be possible to write an Augmentation definition like `lambda img: Augmentation2(Augmentation1(img)[1])`, which ignores the first element of the tuple returned by Augmentation1 (namely, Augmentation1 itself) and operates on the output image.

**Pipeline-level composability**
Since all Augmentations are fundamentally matrix operations on the numpy.array representation of image data, their compositions are (or can be) given by matrix compositions, which gives us some nice mathematical properties. In particular, the associativity of matrix composition lifts to associativity of Augmentation composition, once module-level composability is ensured. This means that `Augmentation3(Augmentation2(Augmentation1(image)))` is the same no matter whether we combine Augmentation3 and Augmentation2 first or combine Augmentation2 and Augmentation1 first.

It gets better: AugmentationSequences are really just wrappers around lists, and so they also compose associatively: running Sequence1 and then Sequence 2 is the same as running `(Sequence1.augmentations + Sequence2.augmentations)`.

[...]

PR #29, opened 2021-07-15
RNG Reproducibility

proofconstruction:
As discussed in #9 , it will be useful for testing purposes to control the degree of nondeterminism in the library.

To accomplish that goal, this PR does the following:
1. Removes all dependencies on `numpy.random` and replaces them with equivalent uses of `random.random` from Python's stdlib.
2. Updates `README.md` to:
    1. note the library's dependency only on `random.random` for random number generation,
    2. provide information about how to set the random seed to control this, and
    3. include links to useful documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -571,6 +571,12 @@ Encoded with quality range of (10, 15)
 ![JPEG Compression Before](images/Augmentations/JPEGCompressionBefore.png)
 ![JPEG Compression After](images/Augmentations/JPEGCompression.png)
 
+# Reproducibility
+This library uses the Python standard library's [random](https://docs.python.org/3/library/random.html) module to generate pseudorandom numbers. If you want to limit nondeterministic results, consider setting the [random seed](https://docs.python.org/3/library/random.html#random.seed) in programs that use the Augraphy library, by including `random.seed(42)` in your scripts. (The number 42 is not required; feel free to choose your own memorable seed when requiring deterministic RNG).
+
+
+Depending on your application, you may also need to set the [numpy random seed](https://docs.scipy.org/doc/numpy-1.15.0/reference/generated/numpy.random.seed.html) or the seeds for other sources of randomness (for a brief overview, see [here](https://pytorch.org/docs/stable/notes/randomness.html)), but Augraphy does not depend on these for random number generation.
+
 # Contributing
 Pull requests are welcome. For major changes, please open an issue first to discuss what you would like to change.
 
--- a/src/augraphy/augmentations/brightnesstexturize.py
+++ b/src/augraphy/augmentations/brightnesstexturize.py
@@ -45,9 +45,9 @@ class BrightnessTexturizeAugmentation(Augmentation):
 
             low = value - (value * self.deviation)  # *random.uniform(0, deviation)
             max = value + (value * self.deviation)
-            brightness_matrix = np.random.uniform(
-                low, max, (hsv.shape[0], hsv.shape[1])
-            )
+
+            makerand = np.vectorize(lambda x: random.uniform(low,max))
+            brightness_matrix = makerand(np.array((hsv.shape[0], hsv.shape[1])))
             hsv[:, :, 1] *= brightness_matrix
             hsv[:, :, 2] *= brightness_matrix
             hsv[:, :, 1][hsv[:, :, 1] > 255] = 255
@@ -59,9 +59,9 @@ class BrightnessTexturizeAugmentation(Augmentation):
 
             low = value - (value * self.deviation)
             max = value + (value * self.deviation)
-            brightness_matrix = np.random.uniform(
-                low, max, (hsv.shape[0], hsv.shape[1])
-            )
+
+            makerand = np.vectorize(lambda x: random.uniform(low,max))
+            brightness_matrix = makerand(np.array((hsv.shape[0], hsv.shape[1])))
             hsv[:, :, 1] *= brightness_matrix
             hsv[:, :, 2] *= brightness_matrix
             hsv[:, :, 1][hsv[:, :, 1] > 255] = 255
--- a/src/augraphy/augmentations/noisetexturize.py
+++ b/src/augraphy/augmentations/noisetexturize.py
@@ -75,7 +75,10 @@ class NoiseTexturizeAugmentation(Augmentation):
         if w == 0:
             w = 1
 
-        result = np.random.normal(mean, sigma, (w, h, 1))
+
+        gaussian = np.vectorize(lambda x: random.gauss(mean,sigma))
+        result = gaussian(np.array((w,h)))
+
         if ratio > 1:
             result = cv2.resize(
                 result, dsize=(width, height), interpolation=cv2.INTER_LINEAR
--- a/src/augraphy/base/oneof.py
+++ b/src/augraphy/base/oneof.py
@@ -1,4 +1,3 @@
-import numpy as np
 import random
 
 from augraphy.base.augmentation import Augmentation
@@ -24,13 +23,8 @@ class OneOf(Augmentation):
     def __call__(self, data, force=False):
         if self.augmentation_probabilities and (force or self.should_run()):
 
-            # Seed the random object.
-            random_state = np.random.RandomState(random.randint(0, 2 ** 32 - 1))
-
             # Randomly selects one Augmentation to apply.
-            augmentation = random_state.choice(
-                self.augmentations, p=self.augmentation_probabilities
-            )
+            augmentation = random.choice(self.augmentations)
 
             # Applies the selected Augmentation.
             augmentation(data, force=True)


jboarman:
Looks great!

kwcckw:
> 
> 
> As discussed in #9 , it will be useful for testing purposes to control the degree of nondeterminism in the library.
> 
> To accomplish that goal, this PR does the following:
> 
>     1. Removes all dependencies on `numpy.random` and replaces them with equivalent uses of `random.random` from Python's stdlib.
> 
>     2. Updates `README.md` to:
>        
>        1. note the library's dependency only on `random.random` for random number generation,
>        2. provide information about how to set the random seed to control this, and
>        3. include links to useful documentation.

Thanks, i will try this out.

kwcckw:
i tried but looks like sometimes the results are not consistent when p is not set as 1. 

This is the modified default pipeline:

    ink_phase = AugmentationSequence([
                                  FoldingAugmentation(p=0.5),
                                  BleedThroughAugmentation(p=0.5),
                                  InkBleedAugmentation(p=0.5),
                                  DustyInkAugmentation(p=0.5),
                                  LowInkBlobsAugmentation(p=0.5),
                                  OneOf([
                                    LowInkRandomLinesAugmentation(use_consistent_lines=False,p=0.5),
                                    LowInkRandomLinesAugmentation(use_consistent_lines=True,p=0.5),
                                    LowInkPeriodicLinesAugmentation(use_consistent_lines=False,p=0.5),
                                    LowInkPeriodicLinesAugmentation(use_consistent_lines=True,p=0.5),
                                  ]),
                                  GaussianBlurAugmentation('ink', p=0.5)
                                ])

    paper_phase = AugmentationSequence([PaperFactory(),
                                        OneOf([
                                        AugmentationSequence([
                                            NoiseTexturizeAugmentation(p=0.5),
                                            BrightnessTexturizeAugmentation(p=0.5),
                                            GaussianBlurAugmentation('paper', [(3,3), (3,5), (5,3), (5,5)],p=0.5),
                                            ]),
                                        AugmentationSequence([
                                            BrightnessTexturizeAugmentation(p=0.5),
                                            NoiseTexturizeAugmentation(p=0.5),
                                            GaussianBlurAugmentation('paper', [(3,3), (3,5), (5,3), (5,5)],p=0.5),
                                        ])]),
                                        BrightnessAugmentation('paper',p=0.5)
                                ])


    post_phase = AugmentationSequence([DirtyRollersAugmentation(p=0.0),
                                    OneOf([
                                            LightingGradientAugmentation(p=0.5),
                                            BrightnessAugmentation('post',p=0.5)
                                        ]),
                                        SubtleNoiseAugmentation(p=0.5),
                                        JpegAugmentation(p=0.5)
                                        ])

                                        
1st trial:
image

2nd trial:
image

The test code:

          import string
          import random
          import cv2
          import numpy as np
          import PIL
          import os
          from pdf417gen import encode, render_image
          random.seed(2)
          
          from augraphy.base.augmentationpipeline import AugraphyPipeline
          from augraphy.default.pipeline import default_augraphy_pipeline
          from augraphy.augmentations import *
          
          
          img = cv2.imread('test_img.png')
          
          pipeline = default_augraphy_pipeline()
          
          
          for i in range(2):
              crappified = pipeline.augment(img)
              dim = crappified["output"].shape
              resized=cv2.resize(crappified["output"], (int(dim[0]/2),int(dim[1]/2)), interpolation = cv2.INTER_AREA)
              cv2.imshow(str(i), resized)
              cv2.waitKey(1)

I think somewhere else there's a section is still using numpy.random, need to further check on this.

proofconstruction:
The `default_augraphy_pipeline()` is very random; most augmentations have probability 0.5 and these are not values chosen from the Mersenne twister sequence used by `random.random`. If we build a new pipeline where **every** augmentation has `p=1` (or `force=True`), and use parameters that prevent the `AugraphyPipeline` from randomly selecting more stuff.

Here's a modified script which defines a new pipeline and restricts more of the probability:
```
import string
import random
import cv2
import numpy as np
import PIL
import os

from augraphy.base.paperfactory import PaperFactory
from augraphy.base.oneof import OneOf
from augraphy.base.augmentationsequence import AugmentationSequence
from augraphy.base.augmentationpipeline import AugraphyPipeline

from augraphy.augmentations.brightness import BrightnessAugmentation
from augraphy.augmentations.brightnesstexturize import BrightnessTexturizeAugmentation
from augraphy.augmentations.dirtyrollers import DirtyRollersAugmentation
from augraphy.augmentations.dustyink import DustyInkAugmentation
from augraphy.augmentations.gaussianblur import GaussianBlurAugmentation
from augraphy.augmentations.inkbleed import InkBleedAugmentation
from augraphy.augmentations.jpeg import JpegAugmentation
from augraphy.augmentations.lightinggradient import LightingGradientAugmentation
from augraphy.augmentations.lowinkblobs import LowInkBlobsAugmentation
from augraphy.augmentations.lowinkperiodiclines import LowInkPeriodicLinesAugmentation
from augraphy.augmentations.lowinkrandomlines import LowInkRandomLinesAugmentation
from augraphy.augmentations.noisetexturize import NoiseTexturizeAugmentation
from augraphy.augmentations.subtlenoise import SubtleNoiseAugmentation

random.seed(2)

img = cv2.imread("test.png")

ink_phase = AugmentationSequence([
                              InkBleedAugmentation(p=1),
                              DustyInkAugmentation(p=1),
                              LowInkBlobsAugmentation(p=1),
                              OneOf([
                                LowInkRandomLinesAugmentation(use_consistent_lines=False, p=1),
                                LowInkRandomLinesAugmentation(use_consistent_lines=True, p=1),
                                LowInkPeriodicLinesAugmentation(use_consistent_lines=False, p=1),
                                LowInkPeriodicLinesAugmentation(use_consistent_lines=True, p=1),
                              ], p=1),
                              GaussianBlurAugmentation('ink', p=1)
                            ])

paper_phase = AugmentationSequence([PaperFactory(),
                                    OneOf([
                                    AugmentationSequence([
                                        NoiseTexturizeAugmentation(p=1),
                                        BrightnessTexturizeAugmentation(p=1),
                                        GaussianBlurAugmentation('paper', [(3,3), (3,5), (5,3), (5,5)], p=1),
                                        ], p=1),
                                    AugmentationSequence([
                                        BrightnessTexturizeAugmentation(p=1),
                                        NoiseTexturizeAugmentation(p=1),
                                        GaussianBlurAugmentation('paper', [(3,3), (3,5), (5,3), (5,5)], p=1),
                                    ])], p=1),
                                    BrightnessAugmentation('paper', p=1)
                            ])


post_phase = AugmentationSequence([DirtyRollersAugmentation(p=0.0),
                                OneOf([
                                        LightingGradientAugmentation(p=1),
                                        BrightnessAugmentation('post', p=1)
                                    ], p=1),
                                    SubtleNoiseAugmentation(p=1),
                                    JpegAugmentation(p=1)
                                    ])

pipeline = AugraphyPipeline(ink_phase,
                            paper_phase,
                            post_phase,
                            ink_color_range=(0,0),
                            paper_color_range=(255,255),
                            rotate_range=(0,0)
                            )


for i in range(10):
    crappified = pipeline.augment(img)
    dim = crappified["output"].shape
    resized=cv2.resize(crappified["output"], (int(dim[0]/2),int(dim[1]/2)), interpolation = cv2.INTER_AREA)
    cv2.imwrite(str(i) + ".jpg", resized)
```

Using this, I was able to produce the same results in 5 separate tests (copied the whole augraphy directory to test1, test2, ..., test5 directories and ran src/test.py in each).

kwcckw:
Yes, with p set to 1, i was able to reproduce the same outcome. 

> most augmentations have probability 0.5 and these are not values chosen from the Mersenne twister sequence used by random.random

Then shouldn't we change all value to use generated value from random.random?

Since with fixed random seed, we should be able to reproduce a same result, and i think reproducibility is very important.

proofconstruction:
> Then shouldn't we change all value to use generated value from random.random?
> 
> Since with fixed random seed, we should be able to reproduce a same result, and i think reproducibility is very important.

We almost do this already, but in the opposite order. Right now, we set `p=0.5` [in every augmentation](https://github.com/sparkfish/augraphy/blob/dev/src/augraphy/base/augmentation.py#L12) and then check if [a random number in the interval [0,1] is <= 0.5](https://github.com/sparkfish/augraphy/blob/dev/src/augraphy/base/augmentation.py#L24). We could turn this around and let the default `p=random.randint(0,1)` and change the `should_run` method to
```
    def should_run(self):
        return random.choice([True, False])
```
or maybe `bool(random.randint(0,1))`, but I don't think this really matters.

We already use random numbers for this, but the numbers are generated far away in the Augmentations class, so it's harder to think about ([other libraries also do this](https://github.com/albumentations-team/albumentations/blob/master/albumentations/augmentations/transforms.py#L125)). This is why in #22 I wanted to remove probability from augmentations and put it into the pipeline instead :smile:

kwcckw:
> This is why in #22 I wanted to remove probability from augmentations and put it into the pipeline instead 😄

i see, but if we put it in the pipeline, are we still able to control `p` of each individual augmentation?

jboarman:
I agree.  We need to be sourcing from the same generator instance sharing the same seed across all calls.  How is this problem managed in the other augmentation libs?

proofconstruction:
> i see, but if we put it in the pipeline, are we still able to control `p` of each individual augmentation?

Yes. We would have to add this functionality but it wouldn’t be too hard. I can do it tomorrow. 

> I agree. We need to be sourcing from the same generator instance sharing the same seed across all calls. How is this problem managed in the other augmentation libs?

Other libs just set the seed like we’ve done above.

If you’re already familiar with how seeding the RNG works, skip the rest of this message, but otherwise it goes like this: if the random seed is set, the numbers generated by every successive call to the random object  _in the same process_  are determined. Essentially, there’s a predefined sequence of numbers which the random object stores a cursor pointing into, and when you set the random seed, you specify the starting position for that cursor (technically, you specify the initial _previous_ value). Every call afterwards returns the element at the cursor location and advances the cursor one position in the sequence. By specifying the seed, we can move the cursor back to the same position in the sequence every time the program is run, and this has the practical effect of generating the same “random” numbers every time.

jboarman:
It sounds like we are just getting confused ... a solid example notebook demonstrating any unexpected results is probably the best place to resolve any confusion.  To me, this sounds like it may just come down to where the seed is being set to guarantee 2 separate runs produce the same result.

kwcckw:
> To me, this sounds like it may just come down to where the seed is being set to guarantee 2 separate runs produce the same result.

Yes, that should be the case, so that we can reproduce same result if needed. But in a single run, we should get random result in a loop:

For example, the code:

        import random
        random.seed(1)
        for i in range(10):
            print(random.random())
            
With seed = 1, we will get the same random result below in every run:

      0.13436424411240122
      0.8474337369372327
      0.763774618976614
      0.2550690257394217
      0.49543508709194095
      0.4494910647887381
      0.651592972722763
      0.7887233511355132
      0.0938595867742349
      0.02834747652200631

PR #35, opened 2021-07-17
Fixed some errors in the code.

kwcckw:
1. Fixed typo in BrightnessTexturizeAugmentation.
2. Fixed  channel mismatch in LowInkBlobsAugmentation.
3. Added evaluation to prevent error when there is no paper image in the paper directory.

### DIFF
--- a/src/augraphy/augmentations/dirtyrollers.py
+++ b/src/augraphy/augmentations/dirtyrollers.py
@@ -66,6 +66,8 @@ class DirtyRollersAugmentation(Augmentation):
                 image = cv2.rotate(image, cv2.ROTATE_90_CLOCKWISE)
 
             image = cv2.cvtColor(image, cv2.COLOR_BGR2GRAY)
+            
+            # where is this transform?
             mask = self.transform(
                 self.create_scanline_mask, image.shape[1], image.shape[0], line_width
             )
--- a/src/augraphy/augmentations/brightnesstexturize.py
+++ b/src/augraphy/augmentations/brightnesstexturize.py
@@ -47,7 +47,7 @@ class BrightnessTexturizeAugmentation(Augmentation):
             max = value + (value * self.deviation)
 
             makerand = np.vectorize(lambda x: random.uniform(low,max))
-            brightness_matrix = makerand(np.array((hsv.shape[0], hsv.shape[1])))
+            brightness_matrix = makerand(np.zeros((hsv.shape[0], hsv.shape[1])))
             hsv[:, :, 1] *= brightness_matrix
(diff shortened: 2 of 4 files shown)


src/augraphy/augmentations/dirtyrollers.py
```diff
@@ -66,6 +66,8 @@ def __call__(self, data, force=False):
                 image = cv2.rotate(image, cv2.ROTATE_90_CLOCKWISE)
 
             image = cv2.cvtColor(image, cv2.COLOR_BGR2GRAY)
+            
+            # where is this transform?
```
kwcckw, line 70: @tbrettallen so we should remove the entire self.transform in dirtyrollers augmentation? Right now i can see it is still in the code, and couldn't locate the definition of self.transform.

proofconstruction: This `transform` method call is left over from before [this commit](https://github.com/sparkfish/augraphy/commit/ea0970932b49288b9893e728d8c4442693b825e4) when it was removed. Probably is why the default pipeline has `DirtyRollersAugmentation(p=0.0)` set in the `post_phase`; it doesn't work anymore. I'll make an issue for this.

proofconstruction (approved):
The changes look good, approved.

Notes:
---
My PR including [commit a7b4e7a](https://github.com/proofconstruction/augraphy/commit/a7b4e7ad9dbf8ec318902fe2119142bd5d4a814d) didn't get merged yet, but I corrected the `np.array` -> `np.zeros` problem there. This bug was introduced (by me! thanks for catching it) as part of removing dependencies on `numpy.random`.

The changes in LowInkBlobs look good, but we're moving that out into the library module at `augmentations/lib.py`, since LowInkBlobs and DustyInk are meant for Letterpress. We can fix this later.

The additions to PaperFactory are also good, but I think we should soon find a better way (throwing exceptions) to deal with these kinds of errors than printing to the console.

DirtyRollers is actually broken right now, see issue #36.

kwcckw:
> The changes in LowInkBlobs look good, but we're moving that out into the library module at `augmentations/lib.py`, since LowInkBlobs and DustyInk are meant for Letterpress. We can fix this later.

Right, i would think move to lib.py is better, since actually i need to use those functions in some other augmentations too.

> 
> The additions to PaperFactory are also good, but I think we should soon find a better way (throwing exceptions) to deal with these kinds of errors than printing to the console.
> 

Yes, probably in the future we can introduce certain level of logging at higher level, but that probably can be done later. Using print should be the simplest approach for now.

Anyway, thanks for the feedback.

PR #41, opened 2021-07-22
Fixed dirty roller, added dirty drum and updated default pipeline.

kwcckw:


proofconstruction:
This is great, nice work

PR #46, opened 2021-08-03
requirements.txt to run test.py

faizan1041:


proofconstruction:
I’m actually working on adding proper tests now, and we’ll have a specific requirements.txt for the tests as well as a general one for the project. Closing this for now, but you’re right that we should have this!

PR #50, opened 2021-08-13
Added draft of bad photocopy augmentation.

kwcckw:


proofconstruction:
Can you show a few examples different images with this effect?

kwcckw:
> 
> 
> Can you show a few examples different images with this effect?

Sorry i forgot to include them just now. So actually i'm working on the demo in notebook too and here's some examples using enron email dataset:

example1
example2

Right now it' still an early draft, but it should be usable now and it can be further improvised later.

proofconstruction:
Is there a nice way to make the corners seem more natural? It's too obvious that there are artificial noise patterns applied there; they overlap very clearly in the darkest area. Otherwise these and the ones in the notebook look pretty good. I'd like to see more noise across the whole page though, with some large darker regions.

Merging the PR, we can continue work on the defaults later.

kwcckw:
> 
> 
> Is there a nice way to make the corners seem more natural? It's too obvious that there are artificial noise patterns applied there; they overlap very clearly in the darkest area. Otherwise these and the ones in the notebook look pretty good. I'd like to see more noise across the whole page though, with some large darker regions.
> 
> Merging the PR, we can continue work on the defaults later.

Yup, that's something i will work on in the next update. Right now they are in patches of rectangle shape, and hence causing it to look not natural. Thanks for the feedback.

PR #57, opened 2021-08-20
Added patch for folding augmentation.

kwcckw:
Basically gradient width and height of folding is based on % of image width and height. If the image size is too small , the width or height range will be the same and causing those errors. We would need 4 unique points for the perspective transform. I added an `if` loop to show warning message and prevent further computation if there is no 4 unique points.

### DIFF
--- a/augraphy/augmentations/folding.py
+++ b/augraphy/augmentations/folding.py
@@ -188,19 +188,26 @@ class Folding(Augmentation):
     # Apply perspective transform 2 times and get single folding effect
     def apply_folding(self, img, ysize, xsize, gradient_width, gradient_height, fold_noise):
          
-        min_fold_x = int(gradient_width[0] * xsize)
-        max_fold_x = int(gradient_width[1] * xsize)
+        min_fold_x = min(np.ceil(gradient_width[0] * xsize),xsize).astype('int')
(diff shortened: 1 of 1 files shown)


proofconstruction:
I tested this and still found errors (see below), but I'm still merging this.

# Bug
 `fold_width_one_side` can be 0 [(line 193)](https://github.com/sparkfish/augraphy/blob/dev/augraphy/augmentations/folding.py#L193) if images are less than 20x20. I didn't see this originally when I found the bug with 10x10 images. If images are larger than 20x20, this doesn't seem to happen, so I won't consider this a problem. We will deal with minimum image sizes some other way, see discussion in #58

kwcckw:
> 
> 
> I tested this and still found errors (see below), but I'm still merging this.
> # Bug
> 
> `fold_width_one_side` can be 0 [(line 193)](https://github.com/sparkfish/augraphy/blob/dev/augraphy/augmentations/folding.py#L193) if images are less than 20x20. I didn't see this originally when I found the bug with 10x10 images. If images are larger than 20x20, this doesn't seem to happen, so I won't consider this a problem. We will deal with minimum image sizes some other way, see discussion in #58

Thanks, you are right. So a better way is to check the folding width:

    if (fold_width_one_side != 0) and (fold_y_shift !=0) :

I will include this in the next update:

image

PR #59, opened 2021-08-22
 Refactored and replaced badphotocopy augmentation &  Bug fixed in folding.

kwcckw:


kwcckw:
Example of augmented image with sparse noise parameters:

image

Example of augmented image with non-sparse noise parameters:
image

kwcckw:
@proofconstruction  Could you advise on this error?

image

The error mentioned those files need to be reformatted, but is there any guidance on how to format those files?

proofconstruction:
This means you should run the [`black` reformat tool](https://github.com/psf/black) on your code and then commit again. If you `pip install black` you can then run `black .` in the project directory to format the new files.

kwcckw:
> 
> 
> This means you should run the [`black` reformat tool](https://github.com/psf/black) on your code and then commit again. If you `pip install black` you can then run `black .` in the project directory to format the new files.

Thanks, i will try it out later.

kwcckw:
@proofconstruction Looks like there's still error in python 3.6, and i think the image size is <30 (xs = 16), so the pytest didn't include the min image size limitation?

image
 
Anyway, i will check again by reinstalling python 3.6 later.

kwcckw:
The previous errors should be resolved with my latest push, please check again.

proofconstruction:
Can you explain the changes to bleedthrough and the low ink lines augmentations? The change to folding seems to fix the problem, good work

kwcckw:
> Can you explain the changes to bleedthrough and the low ink lines augmentations? The change to folding seems to fix the problem, good work

For bleedthrough, there's chance image size and mask size is not the same, and this is causing error in the blending function. The possible cause is the small image size.

For folding, I can see sometimes the input image is in int instead of float or uint8. That will causing error in the warp perspective function so I added code to convert image to float.

For applyBlob , when image is too small, size[1] may smaller than size[0] due to the line of code below:

     size = (1, dim // 2 - 1)

This will causing error in the random.randint function when second number is smaller than first number.

Similarly with low ink random lines, the added code is to prevent second number is smaller than first number in the random.randint function.

For low ink periodic line, I added code to prevent zero division error. This can be happened when image size is too small.

proofconstruction (approved):
Looks good. Nice work on the bug fixes!

PR #67, opened 2021-08-30
Added option to select different hash type in badphotocopy.

kwcckw:


kwcckw:
I added option to select different hash type to emulate different effect.  We may add in more hash type later if needed. 

Examples from existing hash type:
image

Examples from new hash type:
image

proofconstruction:
This looks good and I'll merge it once we get the Github Actions checks to pass. Please sync your fork with the latest commits to the official branch.

To fix this, run these in order:
1. `pip install -r requirements-dev.txt`
2. `pre-commit install`
3. `pre-commit run --all-files`
4. `git commit -m ...`

This will install [`pre-commit`](https://pre-commit.com/) which will make sure we don't commit any code without running `black` etc. After you install `pre-commit`, when you try to `git commit`, it will run the same checks we have in Github Actions, and fix any files that don't match. Then you can commit again and everything should pass.

Let me know if you have any trouble

kwcckw:
> This looks good and I'll merge it once we get the Github Actions checks to pass. Please sync your fork with the latest commits to the official branch.
> 
> To fix this, run these in order:
> 
>     1. `pip install -r requirements-dev.txt`
> 
>     2. `pre-commit install`
> 
>     3. `pre-commit run --all-files`
> 
>     4. `git commit -m ...`



So how long usually it took for each step?

This run for half an hour now, is this normal?
image

proofconstruction:
Depending on the network speed, but it shouldn't take so long.

Once they are all downloaded, they will be cached for future use, so you should only have to download once unless parts are updated.

kwcckw:
> 
> 
> Depending on the network speed, but it shouldn't take so long.
> 
> Once they are all downloaded, they will be cached for future use, so you should only have to download once unless parts are updated.

Alright, i will retry again later, thanks.

kwcckw:
@proofconstruction  I executed those command lines, but looks like facing error to add new commit:

Output from command lines:
image

Error:
image

Could you advise again?

kwcckw:
I closed this pull request and reverted the changes from my end. However, i'm still facing the permission denied issue when i try to re-commit again, do you have any idea?

Or would it be easier if i'm just refork the repo and commit again?

proofconstruction:
I think this may be due to committing from inside GitHub Desktop. Can you try from inside a terminal? ([Windows Terminal](https://github.com/microsoft/terminal) is really good)

kwcckw:
> 
> 
> I think this may be due to committing from inside GitHub Desktop. Can you try from inside a terminal? ([Windows Terminal](https://github.com/microsoft/terminal) is really good)

I tried but looks like it's the same:

image

I guess i will need to refork and manually replace the updates again, thanks for the help.

proofconstruction:
[This stackoverflow thread](https://stackoverflow.com/questions/56974927/permission-denied-trying-to-run-python-on-windows-10) seems to indicate that it's a known bug. Maybe if you [update your PATH variable](https://docs.python.org/3/using/windows.html#setting-envvars)?

I'm not familiar with development on Windows, sorry.

kwcckw:
> 
> 
> [This stackoverflow thread](https://stackoverflow.com/questions/56974927/permission-denied-trying-to-run-python-on-windows-10) seems to indicate that it's a known bug. Maybe if you [update your PATH variable](https://docs.python.org/3/using/windows.html#setting-envvars)?
> 
> I'm not familiar with development on Windows, sorry.

Is alright, i tried that previously , non of that working, and it is only occur on this repo, but not the other repo. I will try to refork and try again, thanks.

PR #69, opened 2021-09-01
re-commit badphotocopy update

kwcckw:


kwcckw:
@proofconstruction  i detached the fork and re-commit those updates but looks like i'm facing a new error in those checks, could you advise again?

image

proofconstruction:
black and the trailing comma hook are reformatting badphotocopy.py, you can try to run `pre-commit run --all-files` locally to fix the file and then commit again. It looks like the change is in the init method of that class; black wants us to use a new line for each parameter, and the pre-commit hook adds a comma character after the last parameter

kwcckw:
> 
> 
> black and the trailing comma hook are reformatting badphotocopy.py, you can try to run `pre-commit run --all-files` locally to fix the file and then commit again. It looks like the change is in the init method of that class; black wants us to use a new line for each parameter, and the pre-commit hook adds a comma character after the last parameter

Thanks, it resolved the error.

proofconstruction:
Good - I added some more info about this [here](https://github.com/sparkfish/augraphy/blob/dev/doc/contributing.md)

PR #70, opened 2021-09-06
Updated binaryThreshold.

kwcckw:


augraphy/augmentations/lib.py
```diff
@@ -148,15 +148,82 @@ def applyBlob(
     return mask
 
 
-def binaryThreshold(img, thresholdType=cv2.THRESH_OTSU):
-    """Converts img to grayscale and applies the given threshold type,
-    or uses Otsu thresholding if no thresholdType is given.
+def binaryThreshold(
+    img,
+    enable_otsu=1,
```
proofconstruction, line 153: We can already pass the threshold type in as `cv2.THRESH_OTSU`, why do we need an extra integer flag to turn this on?

augraphy/augmentations/lib.py
```diff
@@ -148,15 +148,82 @@ def applyBlob(
...
+    # convert to grayscale
     grayscale = cv2.cvtColor(img, cv2.COLOR_BGR2GRAY)
-    ret, thresholded = cv2.threshold(grayscale, 0, 255, thresholdType)
+
+    # otsu method:
+    if enable_otsu:
+        ret, thresholded = cv2.threshold(grayscale, 0, 255, cv2.THRESH_OTSU)
+    # simple method
+    elif enable_simple:
+        simple_methods = [
```
proofconstruction, line 203: I wrote this the original way so we can pass `thresholdType=cv2.THRESH_BINARY` or another type, but default to Otsu. Then we don't have to modify the code to support future thresholding types (if any). If we use this, we will need to keep this list updated with any future changes to OpenCV's thresholding methods.

Can we find a cleaner way to accept the simple methods OR the adaptive methods?

kwcckw:
@proofconstruction Could you check and merge this? Would need this function for the Monochrome method in the Faxify augmentation.

augraphy/augmentations/lib.py
proofconstruction: Thought about this more and I think I agree with what you put here.

augraphy/augmentations/lib.py
proofconstruction: Ignore, see other comment.

proofconstruction (approved):

proofconstruction:
All set

augraphy/augmentations/lib.py
kwcckw: Sorry, I should explain more on this update.

So what i did was set otsu thresholding method as default, and user is having the flexibility to choose a direct thresholding method or adaptive thresholding method too. All these 3 thresholding methods are mutually exclusive, so if all of them are set to true at once, otsu thresholding method will be having the highest priority, followed by simple thresholding method. 

[Update] I just noticed some bug which affecting the priority of those methods, i will include the fix in the next update.

PR #87, opened 2021-09-10
Added Faxify augmentation.

kwcckw:


kwcckw:
Example of current halftone effect:

With invert (left) and without invert(right):
image

kwcckw:
@proofconstruction I'm able to pass all the pre-commit process in my local computer, but looks like I'm facing error here:

image

Do you have any idea on this error?

kwcckw:
Also i think we can remove the Downscale augmentation? Since it is included here too.

proofconstruction:
```shell
Reorder python imports...................................................Failed
- hook id: reorder-python-imports
- exit code: 1
- files were modified by this hook

Reordering imports in augraphy/augmentations/faxify.py

Add trailing commas......................................................Failed
- hook id: add-trailing-comma
- exit code: 1
- files were modified by this hook

Rewriting augraphy/augmentations/faxify.py

black....................................................................Failed
- hook id: black
- files were modified by this hook

reformatted augraphy/augmentations/faxify.py
```

You need to run `pre-commit run --all-files` before committing again. The new file has some style problems.

kwcckw:
> 
> 
> ```shell
> Reorder python imports...................................................Failed
> - hook id: reorder-python-imports
> - exit code: 1
> - files were modified by this hook
> 
> Reordering imports in augraphy/augmentations/faxify.py
> 
> Add trailing commas......................................................Failed
> - hook id: add-trailing-comma
> - exit code: 1
> - files were modified by this hook
> 
> Rewriting augraphy/augmentations/faxify.py
> 
> black....................................................................Failed
> - hook id: black
> - files were modified by this hook
> 
> reformatted augraphy/augmentations/faxify.py
> ```
> 
> You need to run `pre-commit run --all-files` before committing again. The new file has some style problems.

Thanks, it resolved the error, but why I need to run it again? Actually this is the 3rd time since I ran it twice in the previous commit too.

proofconstruction:
When do you run `pre-commit run --all-files` ? The system works like this:

1. you finish coding and you're ready to commit the file
2. you `git add` the file, and `git` stores a copy of the file in its "staging" environment
3. `pre-commit` finds a problem with the file and fixes it
4. the fixed version of the file in part (3) and the version from `git add` in part (2) are now different
5. you `git commit` and `git push`, but the file that goes to GitHub is the version from part (2), which has a problem
6. GitHub Actions runs `pre-commit` which finds a problem with the file, and the checks fail like here

It's best to run `pre-commit run --all-files` as soon as you finish coding, before doing anything else. Then it will fix any style problems first, so you never commit a file that will fail these checks. (Don't worry - I'm still trying to remember to do this too :grin: )

proofconstruction:
Also, can you add an example of this augmentation?

kwcckw:
> 
> 
> When do you run `pre-commit run --all-files` ? The system works like this:
> 
>     1. you finish coding and you're ready to commit the file
> 
>     2. you `git add` the file, and `git` stores a copy of the file in its "staging" environment
> 
>     3. `pre-commit` finds a problem with the file and fixes it
> 
>     4. the fixed version of the file in part (3) and the version from `git add` in part (2) are now different
> 
>     5. you `git commit` and `git push`, but the file that goes to GitHub is the version from part (2), which has a problem
> 
>     6. GitHub Actions runs `pre-commit` which finds a problem with the file, and the checks fail like here
> 
> 
> It's best to run `pre-commit run --all-files` as soon as you finish coding, before doing anything else. Then it will fix any style problems first, so you never commit a file that will fail these checks. (Don't worry - I'm still trying to remember to do this too 😁 )

I see, i will try it in my next update.

kwcckw:
> 
> 
> Also, can you add an example of this augmentation?

Do you mean an example on their usage? Or example output? The output example should be in the comment above.

proofconstruction:
> > Also, can you add an example of this augmentation?
> 
> Do you mean an example on their usage? Or example output? The output example should be in the comment above.

Somehow I completely missed it

PR #88, opened 2021-09-14
Added various hash types.

kwcckw:


kwcckw:
I updated the hash types. where:

hash type 1: Noise starts from the edges of page
image

hash type 2: Discontinuous noises
image

hash type 3: Even noise on the whole page
image

hash type 4: Sparse and little noise
image

hash type 5: Noises with uniform pattern 
image

If the hash type value is not within 1-5, it will be using the default hash type, which is the previous randomize noises.

proofconstruction:
This looks great!

proofconstruction:
What were the values for `noise_density` and `max_iteration` for the hash type 4 pictures?

kwcckw:
> 
> 
> What were the values for `noise_density` and `max_iteration` for the hash type 4 pictures?

It is in the code line 42-51:

        if self.hash_type == 1:
            self.noise_density = (0.01, 0.1)
        elif self.hash_type == 2:
            self.noise_density = (0.3, 0.3)
        elif self.hash_type == 3:
            self.noise_density = (0.01, 0.2)
        elif self.hash_type == 4:
            self.noise_density = (0.01, 0.2)
        elif self.hash_type == 5:
            self.noise_density = (0.1, 0.5)
            
Usually max_iteration won't have much effect, and we need the value in range of 7~9.

proofconstruction:
I've run this a few times, and sometimes `BadPhotoCopy(hash_type=4, p=1)` gets me output like this. How can I make sure I always get the sparse noise?

2021-09-14-194208_2160x3840_scrot

kwcckw:
> 
> 
> I've run this a few times, and sometimes `BadPhotoCopy(hash_type=4, p=1)` gets me output like this. How can I make sure I always get the sparse noise?
> 
> 2021-09-14-194208_2160x3840_scrot

Could you send me your image to check on this? I think it may due to different image size or some other bug, and causing this unexpected output.

proofconstruction:
It's this one https://drive.google.com/file/d/1bH1HVnvsOpVnQuXjjxLoTVJawX0or1l3/view?usp=sharing

kwcckw:
> 
> 
> It's this one https://drive.google.com/file/d/1bH1HVnvsOpVnQuXjjxLoTVJawX0or1l3/view?usp=sharing

Thanks, let me check and i will get back to you again.

PR #90, opened 2021-09-16
Updated badphotography, hash type - 4 

kwcckw:


kwcckw:
I tried with 50 images, all images look good to me, so i think this update should be resolved the issue now.

The output images:
[imgs.zip](https://github.com/sparkfish/augraphy/files/7175697/imgs.zip)

proofconstruction:
These look great, exactly what we need for the archetype reproductions

PR #93, opened 2021-09-19
Updated overlaybuilder.py

kwcckw:


kwcckw:
Not sure why, i'm getting this error:
image

The file is too complex?

proofconstruction:
That's `black` complaining about the `buildOverlay` function having too many if/else branches now, but don't worry about it. For this project, I think we sometimes will need a lot of branching logic, so I made the change for `flake8` to ignore this warning. Try to merge the upstream change and see if it's fixed.

kwcckw:
> 
> 
> That's `black` complaining about the `buildOverlay` function having too many if/else branches now, but don't worry about it. For this project, I think we sometimes will need a lot of branching logic, so I made the change for `flake8` to ignore this warning. Try to merge the upstream change and see if it's fixed.

Great, now it resolved the issue, also i noticed there are several issues still need to be fixed (such as the foreground, background, repeat n times related size issue), you will continue to work on it?

proofconstruction:
Which issues? You can work on it if you want, just let me know

kwcckw:
> 
> 
> Which issues? You can work on it if you want, just let me know

When `ntimes` is too large, it would be over the image size, we need to set some limit on that. Also it should be used within the pipeline? So is it intended to use `ImageOverlay` instead of ` OverlayBuilder` inside the pipeline too?

proofconstruction:
Good catch on the `ntimes`; I think we could throw an exception if `ntimes` multiplied by the foreground dimension is larger than the background dimension.

`ImageOverlay` should be used in the pipeline, but `OverlayBuilder` should be used outside of the pipeline. For some pipelines, we will want to build multiple overlays with `OverlayBuilder` and apply them at different times with `ImageOverlay`, like in this image:

2 - News - Let Us Have Facts - 1962 - dirty


The original document had hole punches, but then it was shrunk down and put on a page that had binder clips and binder rings, before it was scanned. To reproduce this, we should use one pipeline with one overlay to construct the "inner" image with the hole punches, and then a second pipeline to overlay the output of the first pipeline in the right position, then add the overlays for binder clips and binder rings. In all, there would be at least 4 `ImageOverlay`s just for these parts.

kwcckw:
> 
> 
> Good catch on the `ntimes`; I think we could throw an exception if `ntimes` multiplied by the foreground dimension is larger than the background dimension.

I think it would be better if we set ntimes to max value of ntimes before we facing error if the ntimes if too large, so that user will still able to run the code, then we can just display a warning to notify user here.
> 
> `ImageOverlay` should be used in the pipeline, but `OverlayBuilder` should be used outside of the pipeline. For some pipelines, we will want to build multiple overlays with `OverlayBuilder` and apply them at different times with `ImageOverlay`, like in this image:
> 
> 2 - News - Let Us Have Facts - 1962 - dirty
> 
> The original document had hole punches, but then it was shrunk down and put on a page that had binder clips and binder rings, before it was scanned. To reproduce this, we should use one pipeline with one overlay to construct the "inner" image with the hole punches, and then a second pipeline to overlay the output of the first pipeline in the right position, then add the overlays for binder clips and binder rings. In all, there would be at least 4 `ImageOverlay`s just for these parts.

But why don't we just implement `OverlayBuilder` inside the pipeline with more input arguements? Or it will be too complex and lesser flexibility in this way?

proofconstruction:
> I think it would be better if we set ntimes to max value of ntimes before we facing error if the ntimes if too large, so that user will still able to run the code, then we can just display a warning to notify user here.

I don't understand; can you show me some code that will do this?

> But why don't we just implement OverlayBuilder inside the pipeline with more input arguements? Or it will be too complex and lesser flexibility in this way?

I thought that, for more complex overlays (like repeating binder clips) we would want to create these overlays outside the pipeline first, and then apply them inside the pipeline. This keeps what we write in `pipeline = AugraphyPipeline(ink_phase = ... )` easier to follow. I think the overlay creation and application logic should be separate mostly because the pipeline is really about transformations on a single image (the document), and to add building the overlays into the pipeline would make even more complex pipelines, with parts that aren't really about transforming the document.

kwcckw:
> 
> I don't understand; can you show me some code that will do this?
> 

A simpler way would be just limiting the starting coordinate to the size of background, for example:

    yloc = min(yloc + fgHeight + offsetHeight, bgHeight-fgHeight)
    xloc = min(xloc + fgWidth + offsetWidth, bgWidth-fgWidth)
    
Otherwise, we can also break from ntimes whenever `yloc + fgHeight + offsetHeight` or `xloc + fgWidth + offsetWidth` is bigger than image size.

> 
> I thought that, for more complex overlays (like repeating binder clips) we would want to create these overlays outside the pipeline first, and then apply them inside the pipeline. This keeps what we write in `pipeline = AugraphyPipeline(ink_phase = ... )` easier to follow. I think the overlay creation and application logic should be separate mostly because the pipeline is really about transformations on a single image (the document), and to add building the overlays into the pipeline would make even more complex pipelines, with parts that aren't really about transforming the document.

Actually i don't see any problem to include it in the pipeline, we can create a clips/holes augmentation with overlay in ink phase, before any other augmentation and apply the overlay effect first. Since it would be more convenient for the users if they can use the overlay effect in the pipeline too.

proofconstruction:
> A simpler way would be just limiting the starting coordinate to the size of background

> Actually i don't see any problem to include it in the pipeline, we can create a clips/holes augmentation with overlay in ink phase, before any other augmentation and apply the overlay effect first. Since it would be more convenient for the users if they can use the overlay effect in the pipeline too.

Can you add these changes to this PR?

kwcckw:
> 
> 
> > A simpler way would be just limiting the starting coordinate to the size of background
> 
> > Actually i don't see any problem to include it in the pipeline, we can create a clips/holes augmentation with overlay in ink phase, before any other augmentation and apply the overlay effect first. Since it would be more convenient for the users if they can use the overlay effect in the pipeline too.
> 
> Can you add these changes to this PR?

Yea, i'm still working on it to add those limitations on nsteps and offset, i will push it again once it is done.

kwcckw:
With this update, now it shows maximum possible overlays, for example, using large nsteps and large offset:

        fg1 = cv2.imread('binderring.png')
        ob1 = OverlayBuilder(fg1, image_input, 300, "right",1000)
        overlay1 = ob1.buildOverlay()
        
 Output:
image

Also if the offset is too large, it will be at the far end  from the input `edge` direction.

proofconstruction:
Looks good, thanks!

PR #95, opened 2021-09-25
Added BindingsAndFasteners and some other fixes. 

kwcckw:


kwcckw:
Here's the examples from Figshare templates:
image

And here's the examples with some manually created effect:
image

kwcckw:
There's some problem with the usage of `try-except`:

Is it possible to disable this requirement?
image

proofconstruction:
This is great, nice work and thanks for cleaning up my mess :smile: 

> Is it possible to disable this requirement?

This was caused by only having `except:`. This is equivalent to `except Exception:`, which is the more correct way of writing this now [according to flake8](https://www.flake8rules.com/rules/E722.html).

kwcckw:
> 
> 
> This is great, nice work and thanks for cleaning up my mess 😄
> 
> > Is it possible to disable this requirement?
> 
> This was caused by only having `except:`. This is equivalent to `except Exception:`, which is the more correct way of writing this now [according to flake8](https://www.flake8rules.com/rules/E722.html).

Thanks, this should be helpful in case we need to use `try-except` in the code.

Also where do you store those example notebooks for each augmentation? Could you share the folder containing those notebooks?

proofconstruction:
> Also where do you store those example notebooks for each augmentation? Could you share the folder containing those notebooks?

Sure, send me an email to source@proof.construction so I can add you

PR #97, opened 2021-09-29
Added nscales in overlaybuilder, fixed some bugs and added layer input to all augmentations.

kwcckw:


kwcckw:
I have gone through those archetypes and added this commit to resolve some of the requirements. I will include more features in certain augmentations in my next few updates.

kwcckw:
@proofconstruction Could you check and merge this pull request?

proofconstruction:
Looks good, nice work. I was thinking that allowing to specify the layer in every augmentation will add some more flexibility, and I like the idea of varying the foreground size in image overlays

PR #101, opened 2021-10-03
Added Markup class with markup type "underline" and "strikethrough"

### LINKED ISSUE #100: Markup class for strikethrough/underline
Should I create a `Markup`  ( as discussed in this [comment](https://github.com/sparkfish/augraphy/issues/63#issuecomment-919174591)) with the option to select between strikethrough, underline, or any other similar augmentation ?

shaheryar1:
We can remove the strikethrough.py now

### DIFF
--- a/augraphy/augmentations/__init__.py
+++ b/augraphy/augmentations/__init__.py
@@ -18,6 +18,7 @@ from augraphy.augmentations.lightinggradient import LightingGradient
 from augraphy.augmentations.lowinkblobs import LowInkBlobs
 from augraphy.augmentations.lowinkperiodiclines import LowInkPeriodicLines
 from augraphy.augmentations.lowinkrandomlines import LowInkRandomLines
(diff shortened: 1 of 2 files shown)


proofconstruction:
A few points:

1. `self.marktup_type` should be `self.markup_type`
2. can you try to add the highlighter effect from #103 ?
3. when a different type is given other than "strikethrough", the result is not an underline but an overline: 

markup

shaheryar1:
Fixed 1 and 3. 
Started working on 2  but it will take some time. I will create a separate PR for highlight effect

shaheryar1:
> markup

Can you send me the original image which you used here? @proofconstruction

proofconstruction:
Sure, here you go
lorem

And here's the file it was generated with, in case you need it:
[lorem.odt](https://github.com/sparkfish/augraphy/files/7315413/lorem.odt)

Issue #61, opened 2021-08-25
Add "Leaky Ink" Effect Augmentation


kwcckw:
I came across several images with effect looks like ink is leaking into the paper, and usually the effect is at the borders or edges of paper. Probably we can add this effect too?

Example:
![img_301](https://user-images.githubusercontent.com/52521979/130756654-234ed89a-53a7-4ff6-b3d7-82a4121d81e0.png)
![img_137174](https://user-images.githubusercontent.com/52521979/130756669-92a018ae-abe5-4fad-8ca4-4e239357949a.png)

 A more severe ink leaking example:
![img_136996](https://user-images.githubusercontent.com/52521979/130756790-4145fa74-156e-49d5-b8ad-32928d95ebd4.png)

Please suggest a better name for this augmentation too.

proofconstruction:
This effect appears all the time, especially in scans of faxes of old documents.

I think this can probably be a part of BadPhotoCopy, but we can start by developing it separately. For now, we could use BadFax for the name, to fit BadPhotoCopy.

In all of these images, we can see a white border around the text inside the large dark regions. We can probably do something like the following to replicate this effect:
1. randomly select a region to add this effect to,
2. use the Sobel filter to get edges in this region (A),
3. add noise around the edges in this region to "enlarge" the edge mask (B),
4. subtract the edge mask from the region, removing the text
5. apply very heavy noise in the region
6. add the negative of (B) back to the region, creating a white area around the text,
7. add (A) back to the image, printing black text over the white region from part 6

kwcckw:
Right, this should be feasible with current badphotocopy augmentation. Probably we just need to adjust the code a bit to produce a different noise pattern.

proofconstruction:
Can we produce a copy of this with BadPhotoCopy and/or the upcoming Faxify?

kwcckw:
> 
> 
> Can we produce a copy of this with BadPhotoCopy and/or the upcoming Faxify?

Not with Faxify, but it is possible with Badphotocopy, I will work on a different hash type in BadPhotoCopy to produce this effect after my next update on the Faxify.

proofconstruction:
Does the new Faxify update allow this kind of effect?

kwcckw:
> 
> 
> Does the new Faxify update allow this kind of effect?

Nope, faxify is creating something like monochrome (black & white) and halftone. We have something similar but not exactly with badphotocopy (hash type = 1). For example:

[...]

PR #102, opened 2021-10-03
Refactored badphotocopy and added sine pattern in noise with hash type = 1.

kwcckw:


kwcckw:
Here's the example with added sine pattern:

image

So later we can further improve it by using different peak height in the sine wave, that would need some updates in the sine wave equation.

proofconstruction:
This is a great start towards the effect in #61 - but it doesn't look very natural. Maybe if the period of the sine wave is much larger, it might be better.

We have a problem with GitHub though. Can you save the changes for this commit, and then delete your fork and re-fork the repository again? I try to squash merge these PRs to remove these messages: "Merge branch 'dev' of https://github.com/sparkfish/augraphy into dev", but when I squash merge, your branch and the main branch end up out of sync.

proofconstruction:
Also, how do you make the choice for which part to apply the effect to?

kwcckw:
> 
> 
> Also, how do you make the choice for which part to apply the effect to?

I'm checking pixel intensity of a small area at each corner, and then decide `side` of the noise, start from line 344.

Anyway, i will try to undo the commit and commit again, since I would need to contact github support if i need to detach the fork again, that would be more complicated.

kwcckw:
> We have a problem with GitHub though. Can you save the changes for this commit, and then delete your fork and re-fork the repository again? I try to squash merge these PRs to remove these messages: "Merge branch 'dev' of https://github.com/sparkfish/augraphy into dev", but when I squash merge, your branch and the main branch end up out of sync.

I reverted the commit and pushed again the updates. Could you try again？

proofconstruction:
> Anyway, i will try to undo the commit and commit again, since I would need to contact github support if i need to detach the fork again, that would be more complicated.

What I mean is, delete the fork, don't just detach it. Then fork the repo again, and push this commit to the new fork.

Right now your fork is 32 commits ahead, because all of your commits were squashed into one commit and merged into the main repo, but GitHub doesn't keep track of this information. This will only continue to increase, until you make pull requests with hundreds of commits.

To avoid the problem in the future, you can `git rebase` your local commits to squash the "merge branch dev" commit and your actual changes together.

PR #106, opened 2021-10-05
Updates to use chaikin method generate wavy noise.

kwcckw:


kwcckw:
Sorry, looks like i couldn't squash those commits, please try to merge and see if this work, else i will refork again.

With this update, those wave patterns are better now:

left = noise mask, right = output

image

PR #107, opened 2021-10-05
Added "Underline" in markup.py

shaheryar1:


proofconstruction:
It looks like we no longer need the Strikethrough augmentation - what do you think?

shaheryar1:
Yes it can be removed now.

PR #114, opened 2021-10-14
Added noise type in badphotocopy.

kwcckw:


kwcckw:
With this update, we are able to generate some predefined noise patterns. For example:

noise_type = 0  (default : random noise with random setting)
image

noise_type = 1 (noises emerge from the edges of page)
image

noise_type = 2 (stripes of noises)
image

noise_type = 3 (noise spread evenly across the page)
image

noise_type = 4 (sparse and tiny little noise)
image

Also currently `wave_pattern` is still buggy, it should be fixed in the next update.

proofconstruction:
This is great

PR #115, opened 2021-10-16
Updated badphotocopy's  wave pattern

kwcckw:


kwcckw:
With this update, `wave_pattern` in `badphotocopy` should be working now, some examples:

image

PR #119, opened 2021-10-20
Several bugs fixed.

kwcckw:


augraphy/base/augmentationpipeline.py
```diff
@@ -216,6 +211,11 @@ def rotate_image(self, mat, angle):
     def print_ink_to_paper(self, data, overlay, background):
         """Applies the ink layer to the paper layer."""
 
+        # prevent inconsistency in size between background and overlay
+        if overlay.shape[:2] != background.shape[:2]:
+            overlay_y, overlay_x = overlay.shape[:2]
```
kwcckw, line 216: i added this section to prevent size mismatch issue, this is possible when we are having `pageborder` augmentation and it adds several pixels width or height to the output.

kwcckw:
Looks like there's some issues with the test file, i will check again later and fix that.

proofconstruction (approved):
These are much needed changes, good eye!

PR #121, opened 2021-10-24
Added draft of mixed gradients blend method.

kwcckw:


kwcckw:
This is the draft version of mixed gradient blend method, converted from this MATLAB code:
https://github.com/mnkyl/gradient-domain-fusion

At this point , this is the output:

Images:
image

Result when right image as background:
image

Result when left image as background:
image

It is still not working for colour image yet, i will include it in the next update.

proofconstruction:
This looks great!

PR #122, opened 2021-10-31
Updated mixed blend

kwcckw:


kwcckw:
This is the precise mixed blend implementation converted from MATLAB code. I've vectorized all the code and right now there's  left 1 last problem in this section of code (line 624):

    # solve ax=b, where x = a^-1 dot b
    inv_A = inv(csr_matrix(img_A))
    vb = inv_A.dot(b1)
    vg = inv_A.dot(b2)
    vr = inv_A.dot(b3)

In this function, `img_A` is a sparse matrix with size of `m * m` where m = `background row size * background column size `. 

Here, we need to solve `Ax=b`, where `x = A^-1 dot b`

To get the inverse of sparse `img_A`, we need to make it dense but it will easily getting out of memory issue since the size is very large. I've been searching around the web for few days but still didn't getting any possible workaround. Is anyone of you working with large sparse matrix and need to inverse it before? Or do you guys see any other options to tackle on this problem?

Right now it is only working with small size image:

image

kwcckw:
Sorry, looks like this is ax=b is an unsolvable issue, since it is very unlikely to solve a large linear system in an efficient way.

Also I found another workaround which is using ` cv2.seamlessClone`, it's much better and faster. For example:

image

image

I will now close this pull request and i will push again the updates after this.

proofconstruction:
Time complexity for matrix inverse is O(n^2) when n is the size of one side. Large images will cost a *lot* of time to invert, especially when it has to be done twice.

Good idea with `cv2.seamlessClone`, that looks like just what we need

PR #125, opened 2021-11-11
Merged blending into overlaybuilder.

kwcckw:


kwcckw:
This update consolidated blending into current existing `overlaybuilder`.

I added example code inside `overlaybuilder.py` too. So right now there's 3 types of blending method: 

1. min (select minimum value) (The previous blending method in `overlaybuilder`).
2. max (select maximum value)
3. mix (opencv mixed blend)

The examples in the code:
image

Some other examples:
image

proofconstruction:
Good addition -  are the blends always this "jagged" though?

kwcckw:
> 
> 
> Good addition - are the blends always this "jagged" though?

I think it should be case by case. For normal use case, mix blend from opencv should be good enough. Using min only relevant for documents, where ink appears darker. Max blend looks rough in the example, i still cannot think of the use case yet, right now it is included in case we will need it later.

PR #126, opened 2021-11-13
Optimized and simplified Letterpress.

kwcckw:


kwcckw:
In this update, Letterpress is further optimized due to it's slow in the previous version (10 seconds ++ in my machine). Right now it took 1 seconds ++. Those libraries from `lib.py` are no longer needed.

I also simplified the input arguments, here's the explanations to the inputs:

`n_samples` is the number of points in each cluster.

`n_clusters` is the number of noise cluster in the image. 

Take this generated mask of noise as example:
image

Each white spot area is a cluster (controlled by `n_clusters`), while number of points in a cluster is controlled by `n_samples`.

`std_range` control the spreadness of points in a cluster. 

Low `std_range` (value = 100-200):
image

High `std_range` (value = 1000-2000):
image

`value_range` control the value of point, value range from 0-255 (0 = black, 255 = white).

@proofconstruction  I also added usage example at the bottom of the code, i think this should be more convenient, why not we do this for all the other augmentations too?

Here the output from current code's example:
image

kwcckw:
@proofconstruction  Could you check and pull this commit?

proofconstruction:
Great job improving the speed of this augmentation

I like the idea of example code, but in this case nobody will `git clone` the repository and use `python letterpress.py` directly, so the `__main__` method will never execute.

For now, can you move this example to the markdown documentation instead?

kwcckw:
> For now, can you move this example to the markdown documentation instead?

You mean something like this?

image

proofconstruction:
I meant add this to `doc/augmentations/letterpress.py`

kwcckw:
> 
> 
> I meant add this to `doc/augmentations/letterpress.py`

Sure, i add that in the next update.

kwcckw:
In the latest update, i added PIL library alpha blending method to our existing `OverlayBuilder`. All the while `bleedthrough` augmentation is using this alpha blending method and now I've moved the alpha blending section to `OverlayBuilder`.

proofconstruction:
Whoops, closed this accidentally.

Looks good! Thanks for taking care of this

PR #127, opened 2021-11-21
Updates in dirtydrum.

kwcckw:


kwcckw:
In this update, i changed the blending algorithm and noise generation algorithm. In addition, i uncommented those `gaussianblur` related code since it is only applied to the noise mask, but not directly to the actual image, so it shouldn't causing any information lost.

I think the output looks better now, for example:

Actual image:
image

Augmented images:
image

image

image

proofconstruction:
I think these look much more realistic 👍

PR #128, opened 2021-11-28
Added noise_generator.

kwcckw:


kwcckw:
In this update, i added `noise_generator` to move all noise generation related function into 1 single object. So in the next few updates, I will continue to migrate and let those related augmentations to use this `noise_generator` instead. In term of speed, it should be faster or same with current noise generation method.

Up to this point, it is able to generate 8 types of noises:

        noise types:
        1 = single cluster noise 
        2 = noise with regular pattern
        3 = noise at the borders of image
        4 = sparse and little noise
        5 = noise concentrated at left edge
        6 = noise concentrated at right edge
        7 = noise concentrated at top edge
        8 = noise concentrated at bottom edge

image

proofconstruction:
The primary use for this library will be installing it from PyPI and importing it into other scripts, so we don't need to include code that allows people to use each augmentation as a standalone script. Can you remove the part after line 262? The usage example can go in a `doc/augmentations/noise_generator.md` file that documents this augmentation.
 
``` python
# usage example
if __name__ == "__main__":
...
```

kwcckw:
> The primary use for this library will be installing it from PyPI and importing it into other scripts, so we don't need to include code that allows people to use each augmentation as a standalone script. Can you remove the part after line 262? The usage example can go in a `doc/augmentations/noise_generator.md` file that documents this augmentation.
> 
> ```python
> # usage example
> if __name__ == "__main__":
> ...
> ```

Sure, i will remove it in the next update.

doc/augmentations/badphotocopy.md
```diff
@@ -29,3 +14,59 @@ augmentation = BadPhotocopy(
 | `p`             | The probability this augmentation will be applied.               |
 
 Click here for an [example Colab Notebook](https://colab.research.google.com/drive/1_n7atcI-xVSpnV17YuzZhd9ISe735fBo?usp=sharing)
+
+**Example Usage:**
```
kwcckw, line 18: @proofconstruction I changed the structure of doc a bit to move the example code down since the example section will be quite long now.

I will continue to add in more examples for the other augmentations in the next few updates.

proofconstruction: Good idea

PR #129, opened 2021-12-05
Added new overlay methods.

kwcckw:


kwcckw:
Following the latest discussion here
https://github.com/sparkfish/augraphy/pull/124

I added 5 new overlay methods  based on this repo: https://github.com/flrs/blend_modes

Examples:

Background and foreground:
image

Outputs:
image
image
image
image
image

I will include more methods in the next update after i test them out.

augraphy/augmentations/markup.py
```diff
@@ -121,24 +122,42 @@ def __call__(self, image, layer=None, force=False):
...
                             p2,
                             self.markup_color,
                             markup_thickness,
                             lineType=cv2.LINE_AA,
                         )
 
-        if self.markup_type == "highlight":
-            markup_img = cv2.addWeighted(overlay, 0.5, markup_img, 1 - 0.5, 0)
+        # smoothen highlight mask to make it more realistic
+        markup_mask = cv2.GaussianBlur(markup_mask, (7, 7), cv2.BORDER_DEFAULT)
```
kwcckw, line 149: In this update, i added an external mask so that we can further process the mask before applying the overlay process.

For example:
On the left is the mask, while on the right is the overlayed image
image


In addition, I added gaussian blur here too, and i think more relevant filters can be applied here later as well.

For example:
The top image is the one with gaussian blur , while the bottom one is without gaussian blur. I would think it looks more realistic with gaussian blur.
image

kwcckw:
Looks like there's some conflicts, i will check and update it again tomorrow.

augraphy/augmentations/markup.py
jboarman: Indeed ... the gaussian blur definitely makes the edges look more natural.

kwcckw:
> I will include more methods in the next update after i test them out.

From my latest several commits, I added all the overlay methods here: 

https://blend-modes.readthedocs.io/en/latest/visual_examples.html

So it should be completed now.

proofconstruction:
I think these will be very useful, especially for blending stamps and stuff. Great job!

PR #130, opened 2021-12-12
Several improvements and added some examples.

kwcckw:


augraphy/augmentations/bookbinding.py
```diff
@@ -33,16 +34,23 @@ def __repr__(self):
...
-        for i in range(rows):
-            for j in range(cols):
-                dist = math.sqrt((j * j))
-                d = dist + (radius * (1 - math.cos(angle)))
-                new_i = img[i, j] * (dist / (d)) ** 2
-                img_output[i, j, :] = new_i
-        return img_output
+
+        # compute mask of shadow
+        img_dist = np.repeat(np.arange(cols), rows)
```
kwcckw, line 42: In this update, i vectorized the shadow related section here to speed up the computation speed of book binding.

Previously it took 8-9 seconds in my machine and now it took only 2-3 seconds.

proofconstruction: Great work, this will help a lot especially in Colab notebooks

proofconstruction:
Good additions; vectorizing where possible will help keep augraphy fast enough for comfortable use, and I think we may need to start adding longer examples like these.

PR #131, opened 2021-12-14
Edits and added example to brightness and brightnesstexturize

kwcckw:


augraphy/augmentations/brightnesstexturize.py
```diff
@@ -34,37 +34,57 @@ def __repr__(self):
     # Applies the Augmentation to input data.
     def __call__(self, image, layer=None, force=False):
         if force or self.should_run():
-            image = image.copy()
+            image_output = image.copy()
+
+            # for colour image
+            if len(image.shape) > 2:
```
kwcckw, line 40: Added this section so that this augmentation can process grayscale image too.

PR #132, opened 2021-12-14
Upgraded markup.py

shaheryar1:
- Added single word mode
- Added markup repetitions
- Adjust `alpha=1` for strikethrough and underline and `0.5` for highlight

### DIFF
--- a/augraphy/augmentations/markup.py
+++ b/augraphy/augmentations/markup.py
@@ -1,4 +1,5 @@
 import math
+import os
 import random
 from pathlib import Path
 
@@ -8,7 +9,7 @@ import numpy as np
 from augraphy.augmentations.lib import smooth
 from augraphy.base.augmentation import Augmentation
 from augraphy.utilities import *
-import os
+
 
 class Markup(Augmentation):
 
@@ -53,7 +54,7 @@ class Markup(Augmentation):
         self.markup_type = markup_type
         self.markup_color = markup_color
         self.repetitions = repetitions
-        self.single_word_mode=single_word_mode
+        self.single_word_mode = single_word_mode
 
     def __repr__(self):
         return (
@@ -83,7 +84,7 @@ class Markup(Augmentation):
             cv2.THRESH_OTSU | cv2.THRESH_BINARY_INV,
         )
 
-        if self.single_word_mode == False:
+        if self.single_word_mode is False:
             kernel = cv2.getStructuringElement(cv2.MORPH_RECT, (20, 1))
         else:
             kernel = cv2.getStructuringElement(cv2.MORPH_RECT, (5, 1))
--- a/augraphy/augmentations/markup.py
+++ b/augraphy/augmentations/markup.py
@@ -24,6 +24,8 @@ class Markup(Augmentation):
     :type markup_type: string
     :param markup_color: bgr color tuple.
     :type markup_color: tuple of ints
+    :param repetitions: determines how many time a markup should be drawn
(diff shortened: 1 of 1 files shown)


proofconstruction:
Can you show an example picture of the single word mode, and sample code?

augraphy/augmentations/markup.py
```diff
@@ -74,29 +83,46 @@ def __call__(self, image, layer=None, force=False):
...
             choice = random.choice([False, True])
             if choice:
                 x, y, w, h = cv2.boundingRect(cnt)
                 # avoiding too small contours (width less  20% of the image width)
-                if w < int(markup_img.shape[1] / 5):
+                if w < int(markup_img.shape[1] / 10):
                     continue
                 if num_lines == 0:
                     break
+                print(num_lines)
```
kwcckw, line 125: I guess this print is not needed?

kwcckw:
>     * Removed overlaybuilder being applied on strikethrough and underline as well

I checked and overlaybuilder is applied on strikethrough and underline too?

The whole purpose is to draw strikethrough or underline into the `markup_mask` first, then we may need to apply some filters on it before the overlay process. Otherwise, if we apply them directly to the image, those filters will affect the image too, which in some cases they shouldn't.

augraphy/augmentations/markup.py
```diff
@@ -118,35 +127,47 @@ def __call__(self, image, layer=None, force=False):
...
+                                p2,
+                                self.markup_color,
+                                markup_thickness,
+                                lineType=cv2.LINE_AA,
+                            )
 
         # smoothen highlight mask to make it more realistic
-        markup_mask = cv2.GaussianBlur(markup_mask, (7, 7), cv2.BORDER_DEFAULT)
+        if self.markup_type == "highlight":
+            alpha = 0.5
```
kwcckw, line 179: With `alpha` = 0.5, the highlighted colour intensity will be reduced, did you try with 0.5 and 1 and see the differences?

shaheryar1:
> Can you show an example picture of the single word mode, and sample code?

Here are some examples with` single_word_mode=True` and `repetitions=2`
a

a

augraphy/augmentations/markup.py
shaheryar1: Oh I will remove this.

augraphy/augmentations/markup.py
shaheryar1: > The whole purpose is to draw strikethrough or underline into the `markup_mask` first, then we may need to apply some filters on it before the overlay process. Otherwise, if we apply them directly to the image, those filters will affect the image too, which in some cases they shouldn't.

Actually, the overlay is still being applied but for strikethrough and underline the value of alpha is 1 because these markups are mostly drawn with solid colors.

kwcckw: > > The whole purpose is to draw strikethrough or underline into the `markup_mask` first, then we may need to apply some filters on it before the overlay process. Otherwise, if we apply them directly to the image, those filters will affect the image too, which in some cases they shouldn't.
> 
> Actually, the overlay is still being applied but for strikethrough and underline the value of alpha is 1 because these markups are mostly drawn with solid colors.

Yea, actually i meant for highlight, why you set it 0.5? That will reduce intensity of highlighted colour since the alpha value is lower.

kwcckw:
> > Can you show an example picture of the single word mode, and sample code?
> 
> Here are some examples

Looks great.

PR #134, opened 2021-12-16
Edits in dirtyrollers, dustyink and added more examples.

kwcckw:


augraphy/augmentations/dirtyrollers.py
```diff
@@ -12,26 +12,32 @@ class DirtyRollers(Augmentation):
...
 
     # Constructs a string representation of this Augmentation.
     def __repr__(self):
-        return f"DirtyRollers(line_width_range={self.line_width_range}, p={self.p})"
+        return f"DirtyRollers(line_width_range={self.line_width_range}, scanline_type={self.scanline_type}, p={self.p})"
 
     def apply_scanline_mask(self, img, mask, meta_mask):
-        if random.choice([True, False]):
+        # for dark background
+        if self.scanline_type:
```
kwcckw, line 38: I added `scanline_type` argument so that user can choose the proper scanline method. This is due to scanline_mask_v2 doesn't work with white background.

With this equation in `apply_scanline_mask_v2` : `min(255, x + (x * (1 - (y / 100))))`, `x` would be the background and y will be the mask. So if background = white (value = 255), the output will be always 255 too.

For example, with white background:
left = scanline_v2, right = scanline_v1
image

So with white background, user should use scanline_type = 0.

augraphy/augmentations/dustyink.py
```diff
@@ -36,7 +47,8 @@ def __repr__(self):
     def __call__(self, image, layer=None, force=False):
         if force or self.should_run():
             image = image.copy()
-            apply_mask_fn = lambda x, y: y if (x < 64) else x
```
kwcckw, line 39: I added `value_range` so that user is able to choose the min value of pixel to enable the dusty ink effect. Previously it was set at 64, which might not work all the time.

augraphy/augmentations/inkbleed.py
```diff
@@ -18,6 +19,10 @@ class InkBleed(Augmentation):
     :param color_range: Pair of ints determining the range from which color
            noise is sampled.
     :type color_range: tuple, optional
+    :param kernel_size: Kernel size to determine area of inkbleed effect.
```
kwcckw, line 22: I added `kernel_size` so that user is able to adjust the area of inkbleed effect.

For example, 7x7 kernel and 11x11 kernel:
image

augraphy/augmentations/inkbleed.py
```diff
@@ -18,6 +19,10 @@ class InkBleed(Augmentation):
...
+    :type kernel_size: tuple, optional
+    :param severity: Severity to determine concentration of inkbleed effect.
```
kwcckw, line 24: I also added severity, to allow user adjust the concentration inkbleed effect.

For example, severity = (0.1, 0.2) and (0.8,0.9):
image

proofconstruction:
These are all great changes

PR #135, opened 2021-12-24
Added more examples.

kwcckw:


augraphy/augmentations/markup.py
```diff
@@ -150,7 +150,6 @@ def __call__(self, image, layer=None, force=False):
                     # # dividing the line into points
                     points_count = random.randint(3, 10)  # dividing the line into points
                     points = np.linspace(starting_point[0], ending_point[0], points_count)
-                    relative_start_point = starting_point[1] if i % 2 else ending_point[1]
```
kwcckw, line 153: @shaheryar1 This line should be not needed right? I don't see any usage of `relative_start_point`.

augraphy/augmentations/subtlenoise.py
```diff
@@ -22,17 +22,40 @@ def __init__(
     ):
         super().__init__(p=p)
         self.range = range
-        self.add_subtle_noise = np.vectorize(
```
kwcckw, line 25: I removed this section and added vectorized method with numpy (line 36 below). With this, the processing time is reduced to <0.5 seconds from the previous 10 seconds ++.

PR #138, opened 2021-12-30
Added flexibility to rotate image at any phase.

kwcckw:


kwcckw:
In this update, i added the code to enable user rotate image at any phase, this is needed since some of the archetypes would need rotation after the ink phase augmentations. For example:

On the right image, we need to rotate image after we apply the ink phase related augmentations:
image
 
Or would it be better if we just create another rotation based augmentation?

proofconstruction:
I don't think this responsibility belongs in the pipeline class itself, because you end up needing to write a lot of checks for "pre-ink" etc. I do think we should have geometric operations (rotation, shifting, scaling) available as augmentations, because they're so useful and because we want to apply them as steps in a pipeline. So far we've just left this open for people to pull in other libraries like `Albumentations` or `imgaug` to do these kinds of transformations because they aren't strictly document-related, but these transformations are all things that can be accomplished with commodity printing and scanning hardware. I think we could safely add a geometry augmentation to handle all of these.

kwcckw:
> geometry augmentation

Alright, then we can remove all rotation related functions in `agmentationpipeline` and move it to `geometry` augmentation. I will push again later to update on this.

proofconstruction:
Can you add these changes over the existing dev branch, so we don't have the commits adding and removing rotation stuff from the pipeline?

kwcckw:
> Can you add these changes over the existing dev branch, so we don't have the commits adding and removing rotation stuff from the pipeline?

Since we are having the Geometric augmentation, i think the rotation related stuff in pipeline would be redundant now, any reason to not remove them?

kwcckw:
Closed with another pull request later.

PR #140, opened 2022-01-05
Recommit and added more improvements to reproduce archetype.

kwcckw:


proofconstruction:
Another nice addition!

PR #146, opened 2022-01-20
Edits in folding and bookbinding.

kwcckw:


kwcckw:
In this commit, i added mirror effect in bookbinding effect and user input x location for folding effect so that we can generate effect like this:

image

proofconstruction:
This is great!

PR #147, opened 2022-01-22
Edits in bookbinding.

kwcckw:


kwcckw:
With this commit, now bookbinding is able to provide a more realistic bending effect with the added bending part on top of the page (previously it is bend at the bottom of the page only). For example:

Input image:
image

Output:
image

PR #152, opened 2022-01-30
Major changes in NoiseGenerator.

kwcckw:


kwcckw:
In this commit, i restructured NoiseGenerator and now we are able to select side for noise. However, some `noise_type` might not able to select side due to the nature of `noise_type`. For example, noise at all borders (`noise_type` =3) can't select side.

For example, noise mask with predefined side:
image

I also updated `noise_type` = 2 to reproduce noise with some even space in between. Some examples:
image

Issue #150, opened 2022-01-27
Extract pipelines from Colab notebooks


proofconstruction:
Each of the archetype reproduction issues* links to a Colab notebook where development occurred on the pipeline reproducing that archetype. We should take each of these pipelines and make them available as functions in the `augraphy.default` namespace, alongside the `default_augraphy_pipeline`.

* [#76 , #77 , #78 , #79 , #81 , #81 , #82 , #83 , #84 , #85 ]

proofconstruction:
Working on this on my [colab-pipelines](https://github.com/proofconstruction/augraphy/tree/colab-pipelines) branch.

jboarman:
@proofconstruction Can we close this issue?

jboarman:
@kwcckw I'm reopening this issue since I believe you are still working on these.  I saw that you've extracted some of these into the available default pipelines which is great.  I'd also like to ensure we have the notebooks available in an `/examples` folder as a type of documentation that people can open to directly test and play with on their own.

kwcckw:
> @kwcckw I'm reopening this issue since I believe you are still working on these. I saw that you've extracted some of these into the available default pipelines which is great. I'd also like to ensure we have the notebooks available in an `/examples` folder as a type of documentation that people can open to directly test and play with on their own.

Yes, i will include the notebook examples later after i completed all the archetype pipelines. Most of the time some changes will be needed in the augmentations to generate a general pipeline so I'm thinking to add those examples later once we finalized those changes.

PR #154, opened 2022-02-03
Add archetype pipelines, more tests, and make pipelines directly callable

proofconstruction:
This PR:

- adds the archetype pipelines mentioned in #150 to `augraphy.default`
- adds more tests to run those pipelines against random images
- makes pipelines return augmented images directly

### DIFF
--- a/augraphy/default/__init__.py
+++ b/augraphy/default/__init__.py
@@ -1,3 +1,25 @@
-from augraphy.default.pipeline import default_augraphy_pipeline
+from augraphy.default.aktives_und_passives_rauchen import aktives_und_passives_rauchen
+from augraphy.default.default_augraphy_pipeline import default_augraphy_pipeline
+from augraphy.default.diahann_carroll import diahann_carroll
+from augraphy.default.harley_davidson import harley_davidson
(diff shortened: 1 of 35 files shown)


proofconstruction:
Keeping this a draft until we resolve [this](https://github.com/proofconstruction/augraphy/issues/1).

jboarman:
@proofconstruction Do you think this approach to building out a test suite still makes the most sense?

Two directions on testing I see are:

1. **Visual Testing / Image Snapshot Drift** (see [Percy Visual Testing](https://percy.io/visual-testing) as an example) where the rendered output is compared to a baseline image and SSIM or similar metric is used to ensure not too much drift occurs
2.  **Benchmark Testing** where certain activities are measured based on activities performed on the output are tracked to see if the quantitative measures drift, similar to visual testing but focused on task-oriented metrics

IMO, we should probably have a snapshot image for each augmentation run in isolation, and possibly a few in combination, where we track drift from a committed baseline sample image.  Is this basically where you were headed with this branch?

Running notebooks to ensure they don't bomb seems like a valid approach as well, but I think that approach is captured in our _Shabby Page of the Day_ daily build process.

PR #157, opened 2022-02-19
Major improvement in dirty drum effect.

kwcckw:


kwcckw:
With this commit, dirty drum effect should look better now. In addition, user will be having the flexibility to control the effect as below:

`line_concentration` to control amount of dirty drum lines (left - `line_concentration = 0.1`, right - `line_concentration = 0.9`):
image

`noise_intensity` to control amount of noise (left - `noise_intensity = 0.1`, right - `noise_intensity = 0.9`):
image

kwcckw:
With this edit in both lowinklines, user is able to add noise into the line effect. For example:

`noise_probability = 0`, `noise_probability = 0.1` and `noise_probability = 0.9`:
image

kwcckw:
With this update in `BleedThrough` augmentation, now it will be using random images from Figshare library. 

Right now there are 3 selections for `dpi` - 100, 200 and 300. And due to the limitation with Figshare, there's only 500 unique images in the Figshare library.

Some examples:
image

image

image

PR #161, opened 2022-02-26
Major update in PageBorder augmentation.

kwcckw:


kwcckw:
In this update, i added some curvy part to the generated page border lines, it should look more realistic now:

image

image

image

PR #162, opened 2022-03-01
Added feature of logging each augmentation's probability in augmentat…

kwcckw:
…ion pipeline.


kwcckw:
In this update, i added option to enable user log each augmentation running probability.

For example:
pipeline = AugraphyPipeline(ink_phase, paper_phase, post_phase, log_prob=True)

Each run will log a text file in the "./log_prob" folder, and right now it is using hour, minute and second to differentiate each file:
image

In the log file ,each line would be each augmentation name and the running status (True = run, False = norun). For example:
image

With those log files, we can have another script to process and get the overall running statistic, for example, here's the one i'm using now to generate the augmentation running statistics in pdf output:
[process_Log.zip](https://github.com/sparkfish/augraphy/files/8162524/process_Log.zip)

I test with 100 runs (with default pipeline) and here's the output:
[probability_output.pdf](https://github.com/sparkfish/augraphy/files/8162554/probability_output.pdf)

image

Some augmentations with probability = 1 due to they are called with `OneOf` or `AugmentatiuonSequence`, so they will be forced to run when called with those functions. 

From the probability output, they look legit, and i think we need increase the variety of each augmentation default parameter value so that we can see a more variety of augmentations in augmentation output.

augraphy/augmentations/badphotocopy.py
```diff
@@ -43,17 +43,17 @@ class BadPhotoCopy(Augmentation):
...
-        noise_iteration=(1, 1),
-        noise_size=(1, 1),
-        noise_value=(30, 60),
-        noise_sparsity=(0.4, 0.6),
-        noise_concentration=(0.4, 0.6),
-        blur_noise=0,
-        blur_noise_kernel=(5, 5),
-        wave_pattern=0,
-        edge_effect=1,
+        noise_type=None,
```
kwcckw: @proofconstruction Right now i'm setting parameter value to `None` during the class initialization and reinitialize those values again in another function when we call it later (line 265 below). Do you know a better or more elegant way to do this?

kwcckw: This is not relevant now.

PR #167, opened 2022-03-24
Updated brightness and paperfactory.

kwcckw:


kwcckw:
In this update, i added minimum  brightness value in `brightness` augmentation so the outcome will be like this:

left      - texture when not using min brightness value
center - texture when using min brightness value
right    - augmented image using min brightness value

image

image


This should resolve the issue where the paper shadow is blocking the text.

proofconstruction:
Nice. One step closer!

PR #168, opened 2022-03-27
Edits in markup.

kwcckw:


augraphy/augmentations/markup.py
```diff
@@ -166,6 +167,14 @@ def draw_line(self, p1, p2, markup_mask, markup_thickness, reverse):
     def __call__(self, image, layer=None, force=False):
         markup_img = image.copy()
         overlay = markup_img.copy()
+
+        if self.markup_color == "random":
+            self.markup_color = (random.randint(0, 255), random.randint(0, 255), random.randint(0, 255))
```
kwcckw, line 172: In this edit, i added section to generate random colour and the inverted colour (contrast with background). 

For markup, i think it will be more relevant to have it in ink phase, so in that case, most of the time we will not able to detect the colour from paper phase and generate a colour with higher contrast.

kwcckw:
@proofconstruction Do you have any idea on this error? All the checking passed but there's some other error:

image

kwcckw:
@proofconstruction I rechecked on the error in pre-commit and looks like we need new version of black to prevent the error with `_unicodefun`:

https://github.com/psf/black/issues/2964

PR #170, opened 2022-04-12
Added ColorPaper augmentation.

kwcckw:


kwcckw:
In this update, i added `ColorPaper` augmentation to add color in paper based on the user provided hue and saturation value. Some examples:

image

image

image

proofconstruction:
A much-needed addition

PR #177, opened 2022-06-08
Fixed default pipeline.

kwcckw:


augraphy/default/pipeline.py
```diff
@@ -8,7 +8,7 @@
 from augraphy import *
 
 
-def default_augraphy_pipeline(img):
```
kwcckw, line 11: @jboarman @proofconstruction  default pipeline should be just returning pipeline object, not sure why there is an augmentation process within the pipeline now? (line 68 below) 

Or is there any other particular reason for this changes? 

In this commit, i reverted the changes and added an additional function to call default pipeline and augment. `default_augraphy_pipeline` will be just returning the default pipeline now.

PR #189, opened 2022-08-06
Major improvement in bookbinding, edits in pageborder.

kwcckw:


kwcckw:
In this commit, I updated `BookBinding` effect with the addition of `PageBorder` and here are some of the examples:

image
image
image
image

PR #205, opened 2022-11-09
Update reproducibility.md, "sklearn" is deprecated, and update gitignore

proofconstruction:
These commits do the following:
1. add information to `doc/reproducibility.md` about using `random.seed()` to guarantee identical pipelines between Python processes
2. remove `sklearn` and add `scikit-learn` to `requirements.txt` and `requirements-dev.txt` (see [here](https://pypi.org/project/sklearn/))
3. add `augraphy_cache/` to `.gitignore`


doc/reproducibility.md
```diff
@@ -2,3 +2,66 @@
...
+
+The print statements will report whether the shape of the images is the same, and whether there are any non-equal pixels between the two images.
+
+``` bash
+$ python3 testaugmenteds.py
+True
+True
+```
+
+Indeed, the images generated by pipeline executions in successive Python processes are identical, and fixing the random seed made this possible.
```
kwcckw: With this, we just need to set a fixed random seed in the main file before calling augraphy pipeline to enable a consistent output?

proofconstruction: Yes, setting `random.seed(42)` in `myscript.py` will make consistent output between runs of `python myscript.py`

kwcckw: Great, let me test on it first, thanks.

kwcckw: So i just tried it and actually we need to define random seed for all the required functions too. For example, `make_blobs` from `sklearn` is using random seed from numpy, so we need to define the seed number for numpy too.

For safe, i think we can just define like this:
    
    random.seed(42)
    np.random.seed(42)
    cv2.setRNGSeed(42)

jboarman: If we have additional dependencies, it would be better to chain them such that setting the standard Python random seed propagates to these other libs.  In other words, set the seed in these downstream dependencies by simply calling the random.random() to obtain a seed for use in those libs.  Doing this will ensure the RNG output is deterministic regardless of the source.

jboarman: Another complimentary idea Alex had was to include a seed parameter in the pipeline that has a default value, or it could be null if we don't want to set/reset it.  Then we are declaring the method for reproducible randomness within the library's interface so that it is more clear and discoverable.  When the seed is provided, then we will set it per the rules we have documented.

proofconstruction: Commit [2a6c9cd](https://github.com/sparkfish/augraphy/pull/205/commits/2a6c9cd366c4b56061787ecb5032d62cd2086d1b) updates the documentation and adds the seed parameter to the pipeline which in turn sets the `random`, `numpy`, and `cv2` seeds.

kwcckw: > If we have additional dependencies, it would be better to chain them such that setting the standard Python random seed propagates to these other libs. In other words, set the seed in these downstream dependencies by simply calling the random.random() to obtain a seed for use in those libs. Doing this will ensure the RNG output is deterministic regardless of the source.

I think defining the random seed within pipeline is better because it will be quite troublesome if we need to find out each function and define random.random() as their seed (in the case of opencv or numpy). If we missed out some functions, we will need more effort to troubleshoot and search for those functions too.

proofconstruction: > I think defining the random seed within pipeline is better because it will be quite troublesome if we need to find out each function and define random.random() as their seed (in the case of opencv or numpy). If we missed out some functions, we will need more effort to troubleshoot and search for those functions too.

I agree: the seed should be optionally set in the pipeline (see most recent commit), but anyone using individual augmentations outside of the pipeline can still set the seeds manually if they need e.g. reproducible `Letterpress`.

kwcckw: So i think we can merge this pull request?

PR #211, opened 2022-11-20
Edits in lowinklines, badphotocopy and augmentationpipeline.

kwcckw:
1. Fixed bugs in lowinklines.
2. Improve performance of lowinklines.
3. Fixed bugs in badphotocopy's wavepattern.
4. Added evaluation to check for correct image channel in augmentationpipeline.

### DIFF
--- a/augraphy/augmentations/lowinkline.py
+++ b/augraphy/augmentations/lowinkline.py
@@ -32,26 +32,6 @@ class LowInkLine(Augmentation):
         self.use_consistent_lines = use_consistent_lines
         self.noise_probability = noise_probability
 
-        inconsistent_transparency_line = lambda x: random.randint(0, 255)
-        self.inconsistent_transparency_line = np.vectorize(
-            inconsistent_transparency_line,
-        )
-
-        inconsistent_transparency_line_topbottom = (
-            lambda x: random.randint(0, 255) if random.random() > (1 - self.noise_probability) else x
-        )
-        self.inconsistent_transparency_line_topbottom = np.vectorize(
-            inconsistent_transparency_line_topbottom,
-        )
-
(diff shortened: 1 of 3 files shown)


augraphy/augmentations/lowinkline.py
```diff
@@ -32,26 +32,6 @@ def __init__(
         self.use_consistent_lines = use_consistent_lines
         self.noise_probability = noise_probability
 
-        inconsistent_transparency_line = lambda x: random.randint(0, 255)
```
kwcckw, line 35: In this commit, i removed all `np.vectorize` an replace it by np array broadcasting method. Using an image with size of 1200x1200, the speed improved from 1.6 to 0.2 seconds in periodiclines and 4 to 0.2 seconds in randomlines.

augraphy/augmentations/lowinkline.py
```diff
@@ -105,23 +93,56 @@ def add_transparency_line(self, mask, y, alpha=None):
                     )[0]
 
         else:
-            low_ink_line = self.inconsistent_transparency_line(mask[y, :])
+            low_ink_line = (np.random.random((xsize)) * 255).astype("uint8")
+            if len(mask.shape) > 2:
+                new_low_ink_line = np.zeros((1, xsize, mask.shape[2]), dtype="uint8")
```
kwcckw, line 98: In the previous code, noises are added randomly in 3 channels of mask and this created random colour in the created lines:

image

This is fixed in the current commit by assigning all channels with a same noise:
image

augraphy/augmentations/lowinkline.py
jboarman: Wow.  Nice update!

Issue #210, opened 2022-11-18
No module named 'sklearn'


cs-mshah:
I created a separate environment to run an augraphy pipeline on my server, but in spite of having 'sklearn' in the pip list, augraphy is throwing this error:

```
  File "training.py", line 1, in <module>
    from augraphy import *
  File "/home/rivest/anaconda3/envs/ocr/lib/python3.7/site-packages/augraphy/__init__.py", line 1, in <module>
    from augraphy.augmentations import *
  File "/home/rivest/anaconda3/envs/ocr/lib/python3.7/site-packages/augraphy/augmentations/__init__.py", line 1, in <module>
    from augraphy.augmentations.badphotocopy import BadPhotoCopy
  File "/home/rivest/anaconda3/envs/ocr/lib/python3.7/site-packages/augraphy/augmentations/badphotocopy.py", line 6, in <module>
    from augraphy.augmentations.lib import smooth
  File "/home/rivest/anaconda3/envs/ocr/lib/python3.7/site-packages/augraphy/augmentations/lib.py", line 16, in <module>
    from sklearn.datasets import make_blobs
ModuleNotFoundError: No module named 'sklearn'
```

Here is the `pip list` output:
```
(ocr) $ pip list
Package                      Version
---------------------------- --------------------
absl-py                      1.3.0
apturl                       0.5.2
asttokens                    2.1.0
astunparse                   1.6.3
backcall                     0.2.0
bcrypt                       3.1.7
blinker                      1.4
Brlapi                       0.7.0
cachetools                   5.2.0
certifi                      2019.11.28
chardet                      3.0.4
charset-normalizer           2.1.1
Click                        7.0
colorama                     0.4.3
command-not-found            0.3
contourpy                    1.0.5
cryptography                 2.8
cupshelpers                  1.0
cycler                       0.11.0
dbus-python                  1.2.16
debugpy                      1.6.3
decorator                    4.4.2
defer                        1.0.6
distro                       1.4.0
distro-info                  0.23ubuntu1
docker-pycreds               0.4.0
duplicity                    0.8.12.0
entrypoints                  0.3
executing                    1.2.0
fasteners                    0.14.1
flatbuffers                  22.9.24
fonttools                    4.37.4
future                       0.18.2
gast                         0.4.0
gitdb                        4.0.9
GitPython                    3.1.29
google-auth                  2.13.0
google-auth-oauthlib         0.4.6
google-pasta                 0.2.0
grpcio                       1.49.1
[...]

PR #213, opened 2022-12-02
Update setup.py

odulcy-mindee:
Related to https://github.com/sparkfish/augraphy/issues/210.

I was still not able to install `augraphy` using `pip install git+https://github.com/sparkfish/augraphy`. Since 1st December, using the name `sklearn` is deprecated and the build can fail (cf https://pypi.org/project/sklearn/).

This PR should solve this issue.


kwcckw:
Thanks!

PR #220, opened 2023-01-10
Remove hard-coded `True` in `faxify.py`

odulcy-mindee:
I removed an hardcoded `True` in `faxify.py`


proofconstruction:
Good catch!

PR #240, opened 2023-03-02
Added __call__ in augmentation pipeline and updated Function.

kwcckw:


augraphy/utilities/function.py
```diff
@@ -6,25 +6,22 @@ class Function(Augmentation):
     """Accepts an arbitrary function or list of functions to apply in the pipeline.
 
     :param fs: The function(s) to apply.
-    :type fs: function or list of functions
```
jboarman, line 9: @kwcckw Is there a reason that we cannot support a list of functions here as a type of on-the-fly set of augmentations?  I think this was added to support an additional method of composability.  Are there scenarios where this is problematic?  This would be the perfect sort of thing to set up unit tests on, without concern for the output of any augmentation and instead just on the ability to run various pipeline configurations and assert that each augmentation in those pipelines were run as expected.  The augmentations for the purposes of testing the composition would be no-op type augmentations.

jboarman (commented):
See note about the support for function lists

augraphy/utilities/function.py
kwcckw: I think this can be simpler and clearer with just `AugmentationSequence` wrapping multiple `Function`? With this it will be more code efficient and less redundancy. It shouldn't be a problem if we want to revert it back to support list too.

jboarman: As long as supporting the function list here doesn't conflict, then leaving it is similar to the idea of having the `__call__` interface in that is provides ways to do things that require less typing.  Fewer things to read can also increase readability as long as the intent remains clear.

kwcckw: Alright, then i will include it in the next update.

PR #246, opened 2023-03-16
Added padding method in Geometric.

kwcckw:


augraphy/augmentations/geometric.py
```diff
@@ -24,7 +24,13 @@ class Geometric(Augmentation):
            the image rotation.
     :type rotate_range: tuple, optional
     :param randomize: Flag to apply random geometric transformations.
-    :param randomize: int, optional
+    :type randomize: int, optional
+    :param padding: Padding amount on each [left, right, top, bottom] side.
+    :type padding: tuple, optional
+    :param padding_type: Padding methods, select from fill,duplicate and mirror.
```
kwcckw, line 30: So there will be 3 padding methods, `fill` which is fill with any value, `duplicate` is to duplicate the image, including their orientation, and `mirror` which is duplicate but in inversed orientation.

For example:

Fill red value on the left:
image

Duplicate on the left:
image

Mirror on the left:
image

PR #272, opened 2023-04-13
Added new augmentations for Pattern Distortion. #242 

### LINKED ISSUE #242: Add Pattern-based Distortion Generator
Printed documents often include artifacts that may come from the paper textures, printing presses, etc.

The [geopatterns](https://github.com/bryanveloso/geopatterns) python port offers a potential avenue for generating such patterns that could be processed a bit and merged into the paper layer.  Brandon Mills offers a [web-based demo](http://btmills.github.io/geopattern/) of geopatterns.

Below is an illustrative example that comes from the [Tai Le Historical Document Image Binarization Dataset 2021](https://github.com/yddcode/TLHDIBD2021):
image

ss756:
resolves https://github.com/sparkfish/augraphy/issues/242

This PR contains code for pattern-based distortion using 3 different techniques - Voronoi, Delaunay, and Quasicrystals. Perlin Noise is added to the Voronoi and Delaunay Tessellations patterns to create smoother and organic-looking images mentioned in https://github.com/sparkfish/augraphy/issues/242.
Made changes to the docs and benchmark results in README.md


augraphy/augmentations/delaunay.py
```diff
@@ -0,0 +1,230 @@
...
+    :param ws: patch height and width, default value is 200
+    :type ws: int
+    :param p: The probability of applying the augmentation to an input image. Default value is 1.0
+    :type p: float
+
+    """
+
+    def __init__(
+        self,
+        width=500,
```
kwcckw, line 72: @ss756 Are those `width`, `height` and `ws` adjustable? If yes, it will be better to make them into a tuple so that the final value will be selected randomly in the code later. Same with `self.perlin` too. It can be initialized as `random`, so that if `self.perlin` is `random`, the code will randomly set it as True or False.

augraphy/augmentations/delaunay.py
```diff
@@ -0,0 +1,230 @@
...
+        n_vertical_points=(1, 50),
+        perlin=True,
+        ws=200,
+        color_background=(200, 255),
+        p=1,
+    ):
+        super().__init__(p=p)
+        self.width = width
+        self.height = height
+        self.n_points = random.randint(
```
kwcckw, line 85: So we need a different approach here. The actual randomized parameter should be initialized during the `__call__` so that the parameter value will be different from each call.

augraphy/augmentations/delaunay.py
```diff
@@ -0,0 +1,230 @@
...
+                cv2.fillConvexPoly(img, np.array([pt1, pt2, pt3]), color)
+        return img
+
+    # Applies the Augmentation to input data.
+    def __call__(self, image, layer=None, force=False):
+        if force or self.should_run():
+            result = image.copy()
+            h, w, _ = result.shape
+            delaunay_mesh = self.apply_augmentation()
+            # if self.perlin:
```
kwcckw, line 221: Do we still need this section (line 221 to line 224)? Else we can just remove it. If you still need it, please include some related information, maybe something like `# Future development` or something else equivalent.

augraphy/augmentations/delaunay.py
```diff
@@ -0,0 +1,230 @@
...
+                    np.random.randint(min_val, max_val),
+                )
+                cv2.fillConvexPoly(img, np.array([pt1, pt2, pt3]), color)
+        return img
+
+    # Applies the Augmentation to input data.
+    def __call__(self, image, layer=None, force=False):
+        if force or self.should_run():
+            result = image.copy()
+            h, w, _ = result.shape
```
kwcckw, line 219: This line could be just `h, w = result.shape[:2]` because if the input is in grayscale, it will be facing error.

augraphy/augmentations/quasicrystal.py
```diff
@@ -0,0 +1,118 @@
...
+        :param n_rotation: is the number of rotations applied to the pattern, default value lies
+                           between 10 and 15.
+        :type n_rotation: tuple (int)
+
+
+        """
+        super().__init__(p=p)
+        self.imgx = imgx  # width of the image
+        self.imgy = imgy  # hieght of the image
+        self.n_rotation = random.randint(
```
kwcckw, line 67: Similar with the issue i mentioned above, so this should be initialized in the `__call__`.

augraphy/augmentations/quasicrystal.py
```diff
@@ -0,0 +1,118 @@
...
+            n_rotation[0],
+            n_rotation[1],
+        )  # number of rotation to be applied to the pattern
+
+    def __repr__(self):
+        return f"QuasiPattern Distortion: width = {self.imgx} , height = {self.imgy}, n_rotation = {self.n_rotation}"
+
+    def apply_augmentation(self):
+        # Applies the Augmentation to input data.
+        pattern_image = Image.new("RGB", (self.imgx, self.imgy))
```
kwcckw, line 77: Is there any particular reason to use `PIL` instead of `Numpy` here? I think this overall section (line 78 - 106) can be just in `Numpy`?

augraphy/augmentations/quasicrystal.py
```diff
@@ -0,0 +1,118 @@
...
+                for i in range(self.n_rotation):
+                    r = math.hypot(x, y)  # distance between the point to the origin
+                    a = (
+                        math.atan2(y, x) + i * math.pi * 2.0 / self.n_rotation
+                    )  # angle the point makes to the origin plus rotation angle
+                    z += math.cos(r * math.sin(a) * frequency + phase)  # function of cosine added as an offet
+                c = int(round(255 * z / self.n_rotation))  # color
+                pixels[kx, ky] = (c, c, c)  # # RGB value
+        # pattern_image = ImageOps.autocontrast(pattern_image) # increasing the contrast of the image
+        pattern_image.save(
```
kwcckw, line 98: Actually why we need to save the image and load the image again here?

augraphy/augmentations/voronoi.py
```diff
@@ -0,0 +1,180 @@
...
+        img_array = self.generate_voronoi(
+            self.width,
+            self.height,
+            self.num_cells,
+            nsize,
+            (nx, ny, ng),
+            (perlin_x, perlin_y),
+        )
+        image = Image.fromarray(img_array)
+        image.save("images/Voronoi_example.png", "PNG", dpi=(300, 300))
```
kwcckw, line 163: This should be same with `quasicrystal`, why we need to save and load it again?

augraphy/utilities/slidingwindow.py
```diff
@@ -0,0 +1,144 @@
...
+        ****************************************
+
+        :param image: numpy array, original image
+        :type image:  numpy array(int/float)
+        :param mesh_image: numpy array, mesh image
+        :type mesh_image: numpy array(int/float)
+        :param window_size: height and width of patch, default value is 200
+        :type window_size: int
+        """
+        h, w, _ = image.shape
```
kwcckw, line 88: This line should be replaced by `h, w =image.shape[:2]`.

augraphy/utilities/slidingwindow.py
```diff
@@ -0,0 +1,144 @@
...
+        image = cv2.copyMakeBorder(
+            image,
+            top,
+            bottom,
+            left,
+            right,
+            cv2.BORDER_CONSTANT,
+            value=[255, 255, 255],
+        )
+        h, w, _ = image.shape
```
kwcckw, line 101: This line should be replaced by `h, w =image.shape[:2]`.

augraphy/utilities/slidingwindow.py
```diff
@@ -0,0 +1,144 @@
...
+                and wx > 0
+                and initial_coords[0] + wx < w
+                and initial_coords[1] + hy < h
+                and initial_coords[0] + wx > 0
+                and initial_coords[1] + hy > 0
+            ):
+                image[
+                    initial_coords[1] : initial_coords[1] + hy,
+                    initial_coords[0] : initial_coords[0] + wx,
+                    :,
```
kwcckw, line 126: We can just remove the `:` here, else it will facing error if input is grayscale. same with line 131 below.

augraphy/augmentations/voronoi.py
```diff
@@ -0,0 +1,180 @@
...
+        mesh = cv2.imread("images/Voronoi_example.png")
+        os.remove("images/Voronoi_example.png")
+        return mesh
+
+    # Applies the Augmentation to input data.
+    def __call__(self, image, layer=None, force=False):
+        if force or self.should_run():
+            result = image.copy()
+            h, w, _ = result.shape
```
kwcckw, line 172: This line should be replaced by `h, w = result.shape[:2]`, and line 179 below can be just :

    result = result[self.ws : h + self.ws, self.ws : w + self.ws]

We might facing error with the additional `:` if the input is in grayscale. So you will need to test the augmentation with grayscale  image too.

augraphy/augmentations/quasicrystal.py
ss756: Yeah I have written a comment in Voronoi.py
We save it as a 300 dpi image and load it

augraphy/augmentations/delaunay.py
ss756: Yes I will integrate this today

augraphy/augmentations/voronoi.py
ss756: Hi @kwcckw  I went through the link you sent. As per my understanding, they are also doing the same thing by saving the image as 300 dpi and reading it from memory.  Let me know if I have misunderstood it so that I can make the changes in the new PR
image

PR #276, opened 2023-04-19
Updates in Pattern Distortion Pipeline 

### LINKED ISSUE #242: Add Pattern-based Distortion Generator

ss756:
resolves #242

This PR contains updates for Pattern Distortion pipeline, such that it can handle gray-scale images as well. The parameters of  each function is randomized in __call__() instead of __init()__ function, so that it has different values in each run. 
Made changes to the docs.


augraphy/augmentations/delaunay.py
```diff
@@ -179,52 +162,140 @@ def apply_augmentation(self):
...
+            self.n_vertical_points = random.randint(
+                self.n_vertical_points_range[0],
+                self.n_vertical_points_range[1],
+            )  # randonly selecting the edge vertical points in the geometric plane
+            if self.noise_type == "random":
+                self.perlin = random.choice(
+                    [True, False],
+                )  # randomly select to apply Perlin Noise on top of the Tessellation
+            else:
+                self.perlin = False
```
kwcckw, line 281: I think here should be `self.perlin` = `self.noise_type` because if user set True on perlin, we need to get it here.

augraphy/augmentations/voronoi.py
```diff
@@ -169,12 +157,52 @@ def apply_augmentation(self):
     def __call__(self, image, layer=None, force=False):
         if force or self.should_run():
             result = image.copy()
-            h, w, _ = result.shape
+            h, w = result.shape[:2]
+            if self.noise_type == "random":
+                self.perlin = random.choice([True, False])
+            else:
+                self.perlin = False
```
kwcckw, line 170: This is similar with the issue i mentioned above.

augraphy/augmentations/delaunay.py
ss756: Hi @kwcckw, the only option we are giving the user is to select the noise_type. It will either be None or "random" for now. If it is random, then Perlin Noise may or may not be added. It is None or any other value, Perlin Noise is not added.

Explicitly the user cannot select True for Noise_type. What I would suggest is that the user selects noise_type = "perlin" so that perlin noise is integrated. Is this ok ?

kwcckw: Yes, this should work, so you need to mention this in the class descriptions too.

ss756: Yes @kwcckw i will add this to the class description

PR #278, opened 2023-04-22
Major updates in Scribbles, renamed to Scribbles and some bugs fixed.

### LINKED ISSUE #53: Add handwritten `Scribbles` in pen, pencil or marker
It could be useful to reproduce random scratches, artifacts or annotations created using a pen or pencil.  Below are sample references for lines created using handwriting instruments.

### Sample References

image
image
image

kwcckw:


kwcckw:
With this update,  now `Scribbles` is able to generate 3 kinds of ink effect:

1. Marker
image

2. Pen
image

3. Pencil
image

jboarman:
Nice update @kwcckw.

Can you share an example that is higher resolution to really see the effects better?

Also, I'm curious to see examples with skeletonize enabled and one without that enabled.

jboarman:
For pencils, the color can [definitely be more than black](https://github.com/kwcckw/augraphy/blob/e47f912e1e866c257e79b6822fa4d4bdd88777ef/augraphy/augmentations/scribbles.py#L182), but I agree that we would tend towards black color for pencil.  It could be color pencils, most often red like a teacher's red pencil.

image

jboarman:
For the marker, transparency is more related to highlighters than markers since felt-tip or permanent markers, for example, tend to be non-transparent similar to the second set of strokes in the examples at the top of #53.  Perhaps we should also add a "highlighter" to the list of stroke types (we could use a similar set of colors as suggested in #53)?

Also, looking at those marker stroke examples, I think we could introduce some level of noise, though it would definitely be different than pencil noise as it would tend to be darker generally and more linear like these examples.  Note that we are missing this effect in the current highlighter augmetentation to make the effect much more realistic there as well.

image

image

### Highlighter example
image

jboarman:
Here's a [sample collab](https://colab.research.google.com/drive/1v-841bazZDxI7Hg9QiuTrvyBrgg7UXmW?usp=sharing) I put together that demos one approach to stroke noise.  I think this can be generalized to a vector-based and more performant manner, but it may take some thought to think that through...
image

kwcckw:
> Nice update @kwcckw.
> 
> Can you share an example that is higher resolution to really see the effects better?
> 
> Also, I'm curious to see examples with skeletonize enabled and one without that enabled.

Here's the example with and without skeletonization:

With skeletonize:
image

Without skeletonize:
image

So for pencil, it looks better with skeletonization because some fonts are thicker, it doesn't looks realistic in that way.

kwcckw:
> For pencils, the color can [definitely be more than black](https://github.com/kwcckw/augraphy/blob/e47f912e1e866c257e79b6822fa4d4bdd88777ef/augraphy/augmentations/scribbles.py#L182), but I agree that we would tend towards black color for pencil. It could be color pencils, most often red like a teacher's red pencil.
> 
> image

Sure, i will perform some experiments first and include them later.

kwcckw:
> For the marker, transparency is more related to highlighters than markers since felt-tip or permanent markers, for example, tend to be non-transparent similar to the second set of strokes in the examples at the top of #53. Perhaps we should also add a "highlighter" to the list of stroke types (we could use a similar set of colors as suggested in #53)?
> 
> Also, looking at those marker stroke examples, I think we could introduce some level of noise, though it would definitely be different than pencil noise as it would tend to be darker generally and more linear like these examples. Note that we are missing this effect in the current highlighter augmetentation to make the effect much more realistic there as well.
> 
> image
> 
> image
> ### Highlighter example
> 
> image

Thanks, i think this is a good add-on on top of current existing inks. I will include this in the next update.  So for `Markup`, do you think it is relevant to include marker too?

jboarman:
> Thanks, i think this is a good add-on on top of current existing inks. I will include this in the next update. So for Markup, do you think it is relevant to include marker too?

Yes, I think `pen`, `pencil` and `marker` should be added to `Markup`.  Normally, `marker` would be too extreme so I'd leave that out of any "random" rotation.  There could be use cases where words need to be redacted where marker could be useful, but again it would not be the typical use.

kwcckw:
> Yes, I think `pen`, `pencil` and `marker` should be added to `Markup`. Normally, `marker` would be too extreme so I'd leave that out of any "random" rotation. There could be use cases where words need to be redacted where marker could be useful, but again it would not be the typical use.

Okay, probably we will need something like `ink_generator` now to generate different ink effect in across augmentations. Initially, those inks could be pencil, pen, marker and highlighter.

PR #284, opened 2023-05-04
Various updates

kwcckw:
1. Major updates in Markup. Markup uses InkGenerator to generate effect now.
2. Improvement and bugs fixed in InkGenerator.
3. Updates in doc.


kwcckw:
With this updates, now `Markup` uses `InkGenerator` and we will be able to generate `Markup` effect by using different inks, for example:

crossed with marker:
image

strikethrough with pencil:
image

underline with highlighter:
image

PR #299, opened 2023-05-31
Major updates in PageBorder.

### LINKED ISSUE #297: Improve PageBorder effect
Right now `PageBorder` effect is drawing lines to create a border like effect, which may not realistic enough. We a better method to create a better effect. Some real examples:

image
image
image

kwcckw:


kwcckw:
With this update, now `PageBorder` create border effect by stacking multiple images isntead of drawing lines, for example:

image
image
image

PR #305, opened 2023-06-10
Added InkColorSwap and InkMottling.

### LINKED ISSUE #303: Add `InkColorSwap` to Replace the Color Used for Lettering in a Document
Historical documents, as just one scenario, often have text that is faded or otherwise have low contrast compared to the background.  Since it can be difficult to source documents that meet this criteria, we need an augmentation that can be used to render these scenarios.

### Effect of `InkColorSwap` Augmentation

* Randomly select contours of glyphs/letters
* Randomly extend the contour selection to neighboring glyphs (horizontally-close contours) that approximate selecting words
* Swap the color of the contours to a randomly-fixed color for the whole document _or_ a random color for each contour

# Reference Images
Note the effect of the red and black text in examples below.  The examples illustrate these effects where (1) the red ink has a lower contrast than the black ink compared to the background color, (2) the red ink has faded more than the black over time, and (3) when converted to greyscale the red ink almost disappears since it lacks the color contrast to differentiate the ink from the ~paper~ papyrus.

## Red / Black Ink (crisp, but low contrast compared to background)
image

## Red Ink Faded More than Black Ink
image

## Greyscale Version Illustrating Impact of Low Contrast Colors
image

### LINKED ISSUE #304: Add `InkMottling` Augmentation to Ensure Ink is Non-Uniform
To render real-world documents, ink sourced from a perfect PDF source should not render as perfect black, for example.  Such examples are _easily_ denoised with threshold filters.  So we need to add noise to the incoming ink layer.

## Effects of `InkMottling` Augmentation
* [ ] The existing ink layer is adjusted by a noise layer
* [ ] The noise could be composed of positive and negative numbers (useful when ink is not strictly black or white) so that levels are adjusted both up and down
* [ ] Noise options would start with fast noise options like `Poisson` or `Gaussian` or `Speckle` noise  ([see deprecated code from skimage](https://github.com/scikit-image/scikit-image/blob/359d4e360d3a6dc489c45a2a33c6bd5fe6b3908d/skimage/util/noise.py#L42-L231))
* [ ] The range of noise values effecting ink should be parameterized -- by default, this range should be low and subtle, often un-noticeable to the human eye unless you are looking for it
* [ ] The noise layer should randomly apply to either all channels, or be independent noise applied separately to each channel (to simulate black rendered by color ink)

## Example Problem Image (what this augmentation will fix)
image

## Example Real-World Mottling of Black Ink
image

kwcckw:


kwcckw:
Some exampels of InkColorSwap:

image
image
image

kwcckw:
Some examples of InkMottling:

image
image
image

PR #315, opened 2023-06-23
Major updates in BookBinding.

### LINKED ISSUE #169: Improved Book Binding Effect
Some noted features that we can learn from [carolineminuscule-groundtruth examples](https://github.com/rescribe/carolineminuscule-groundtruth/blob/master/bsb00046285/bsb00046285.0011.jpeg):

* Opposing curled page could reveal another page of content
* Opposing curled page could have multiple pages
* Page edging can be more separated and slightly wavy

image

kwcckw:
- change to algorithm in BookBinding so that it's better to understand.
- added shadow effect in BookBinding.
- added feature to load cache image to create the mirror side of the BookBinding effect.


kwcckw:
After the updates, some example images:

image
image
image

PR #316, opened 2023-06-27
Added Hollow.

kwcckw:


kwcckw:
Some examples of hollow effect:

Input and augmented:
example 1:
image 
image

example 2:
image
image

example 3:
image
image

PR #317, opened 2023-06-28
Added NoisyLines.

kwcckw:


kwcckw:
Some examples of `NoisyLines`:

With red color lines and large space interval in each lines:
image

With no space interval in lines:
image

With 50% random noise in lines:
image

PR #318, opened 2023-07-01
Added Squish and updated NoisyLines.

kwcckw:


kwcckw:
Some examples of `Squish` augmentation's output:

Without lines:
image

With lines:
image

In both horizontal and vertical direction:
image

PR #319, opened 2023-07-04
Added SectionShift and fixed bug in NoiseGenerator.

kwcckw:


kwcckw:
Some examples of `SectionShift`:

Without filling value (remained the image original value):
image

Filled with black color:
image

Filled with white color:
image

PR #321, opened 2023-07-06
Added DotMatrix.

kwcckw:


kwcckw:
Some examples of `DotMatrix`:

With "triangle" shape:
image

With "rectangle" shape:
image

With "diamond" shape:
image

PR #322, opened 2023-07-09
Major updates in DotMatrix.

kwcckw:


kwcckw:
The updated effect of `DotMatrix`:

image
image
image

PR #323, opened 2023-07-13
Major updates in PageBorder and etc.

kwcckw:
1. Improved the algorithm in PageBorder to create a better rotated page effect in different angles.
2. Added function to rotate image with any angle in PIL.
3. Fixed bugs in cache image loading function.


kwcckw:
Some examples with the updated `PageBorder`:

Page border effect with random angle of stacked pages, and pages are loaded from cache (same page border = 0):
image

Page border effect with random angle of stacked pages, and pages are loaded from cache (same page border = 1):
image
When same page border is True, their dimension and coordinates are matched with centroid.

Page border effect with large angle of rotations (-360, 360):
image

Page border effect with random and small angle rotation:
image

Page border effect with consistent and small angle rotation:
image

PR #324, opened 2023-07-16
Major updates in PageBorder and BookBinding.

kwcckw:
- added trimming parameter in PageBorder to trim any side of added borders (this will be needed in BookBinding).
- added more parameters in BookBinding to separate the parameters for left and right image.


kwcckw:
Some examples of updated `BookBinding`:

with curve down, using cache images and random pages angle:
image

with curve up, using cache images and random pages angle:
image

with more bending on the left image:
image

with more bending on the right image:
image

with equal bending on both pages:
image

PR #325, opened 2023-07-18
Added shadow effect in BookBinding.

kwcckw:


kwcckw:
Some examples of added shadow effect:

image

image

image

PR #326, opened 2023-07-21
Fixed bugs in dotmatrix, inkshifter, sectionshift and voronoi.

kwcckw:


augraphy/augmentations/inkshifter.py
```diff
@@ -95,7 +95,36 @@ def noise_map(self, shape, res=(64, 64)):
         d0, d1 = shape[0] // res[0], shape[1] // res[1]
         angles = 2 * np.pi * np.random.rand(res[0] + 1, res[1] + 1)
         grad = np.dstack((np.cos(angles), np.sin(angles)))
+        gysize, gxsize = grad.shape[:2]
         grid = np.mgrid[: res[0] : res[0] / shape[0], : res[1] : res[1] / shape[1]].transpose(1, 2, 0) % 1
```
kwcckw, line 99: @ss756 I found an issue here where `grid` size doesn't match with transformed `grad` size, and this is mostly due to the `np.ceil` in line 93 above. I added the evaluations below to prune or pad their size so that their size are always consistent. If you have a better solution, please push the update too.

augraphy/augmentations/voronoi.py
```diff
@@ -150,11 +150,15 @@ def apply_augmentation(self):
...
-        os.remove("images/Voronoi_example.png")
+        # try if it is able to save and read image properly, might facing permission issue
+        try:
+            image = Image.fromarray(img_array)
+            image.save(os.getcwd() + "/Voronoi_example.png", "PNG", dpi=(300, 300))
+            # reads it in a format so that it can be applied as a background pattern to the original image
+            mesh = cv2.imread(os.getcwd() + "/Voronoi_example.png")
+            os.remove(os.getcwd() + "/Voronoi_example.png")
+        except Exception:
+            mesh = img_array
```
kwcckw, line 161: @ss756  So there is a problem with the old code here because the folder `images` might not exist in the directory. I removed that directory and change it into a try-except in case there's any permission issue while writing the file. If there is any error, it will just use the `img_array` instead.

PR #327, opened 2023-07-22
Major updates in Folding.

kwcckw:


kwcckw:
With the new updates in `Folding`, now we can apply the folding effect in any angle, some examples:

image
image
image

PR #328, opened 2023-07-25
Major updates in BindingsAndFasteners.

kwcckw:


kwcckw:
With the latest update, a new binder `triangle_clips` is added to the binder list.

Some examples:

triangle clips:
image

When edge offset is 0, it shows the clipped effect instead:
image

clips:
image

When edge offset is 0, it shows the clipped effect instead:
image

PR #330, opened 2023-07-27
Fixed bugs in BrightnessTexturize and updated delaunay.

kwcckw:
1. Fixed the bug in BrightnessTexturize where the color of the image is changed after the augmentation.
2. Added color changing feature in DelaunayTessellation.


kwcckw:
The existing `DelaunayTessellation`  gives only blue color effect, for example:
image

With this update, users will be able to change the color, for example, random colors:
image
image
image

PR #331, opened 2023-07-28
Updated PatternGenerator.

kwcckw:
1. Added feature to change color in PatternGenerator.
2. Updated example image for the NoiseTexturize's doc image.


kwcckw:
Some examples of updated `PatternGenerator`  augmented images (with different colors):

image
image
image

PR #335, opened 2023-08-01
Fixed the augmentation effect in NoiseTexturize.

kwcckw:


kwcckw:
In this update, `NoiseTexturize` is updated to emulate a correct paper texture surface effect, some examples:

image
image
image
image

PR #339, opened 2023-08-05
Added new benchmark results and added new noise type in BadPhotoCopy.

kwcckw:


kwcckw:
In this update, a new Gaussian noise type is added in `BadPhotoCopy`. Some examples:

image
image
image

PR #341, opened 2023-08-08
Added new Worley noise type in BadPhotoCopy.

kwcckw:


kwcckw:
Some examples of new Worley noise effect:

image
image
image

PR #342, opened 2023-08-10
Updated NoiseGenerator.

kwcckw:
- Added noise type using sklearn make_blobs.
- Added noise type with consistent square pattern.
- In the process of restructuring NoiseGenerator.


kwcckw:
In this update, noise type with `sklearn's make_blobs` and noise type with consistent square pattern are added. `NoiseGenerator` and `BadPhotoCopy` will be restructured later and hence we need to assign them each with unique noise type. 

Some examples:

Noise with `sklearn's make_blobs`:
image
image
image


Noise with consistent square pattern:
image
image
image

PR #345, opened 2023-08-14
Fixed bug in Bookbinding and updated default pipeline.

kwcckw:


kwcckw:
Some examples of updated default pipeline augmented images:

image
image
image
image
image
image
image
image
image
image

PR #355, opened 2023-08-30
Added keypoints support in Folding.

kwcckw:


kwcckw:
An example of Folding augmentation output, with additional inputs of alpha layer, mask and keypoints:

Input and output
image

Input alpha:
image

Output alpha:
image

Input mask:
image

Output mask:
image

Input keypoints:
image

Output keypoints:
image

PR #356, opened 2023-08-31
Added bounding boxes support in Folding and fixed bug.

kwcckw:


kwcckw:
An example of bounding boxes input and output with `Folding` augmentation:

Input and output images
image

Input boxes:
image

Output boxes:
image

PR #385, opened 2023-10-24
Added code to generate new texture by quilting image.

kwcckw:


kwcckw:
Image quilting will be useful in generating repeating textures only, some examples:

image
image
image
image
image

PR #387, opened 2023-10-26
Added an initial version of edge based texture in PaperFactory.

kwcckw:


kwcckw:
Some initial examples:

image
image
image
image

PR #388, opened 2023-10-30
Updated to a better edge texture.

kwcckw:


kwcckw:
Some examples of the updated edge texture:

mask of edge texture (top), applied edge texture (bottom)

Example1:
image

Example2:
image

Example3:
image

PR #390, opened 2023-11-01
Added strange pattern and the masking of edge texture.

kwcckw:


kwcckw:
Some examples with the updated edges based texture mask:

image-1
image
image-1

PR #393, opened 2023-11-06
Added light granular texture in PaperFactory.

kwcckw:


kwcckw:
Some examples of light granular texture:

image
image
image

PR #394, opened 2023-11-07
Added curves based edge texture in PaperFactory.

kwcckw:


kwcckw:
Some examples of curvy edge texture with granular texture:

image
image
image

PR #392, opened 2023-11-04
Updated FFT generated texture.

kwcckw:

PR #395, opened 2023-11-08
Added new fine stains texture in PaperFactory.

kwcckw:


kwcckw:
The examples of current fine stains texture:

image
image
image

Where the prior update:https://github.com/sparkfish/augraphy/pull/392 produces a rough version of stains texture:
image
image
image

PR #397, opened 2023-11-13
Added severe stains texture in PaperFactory.

kwcckw:


kwcckw:
Some examples of severe stains texture:

image
image
image

PR #398, opened 2023-11-14

kwcckw:


kwcckw:
Some examples on the new granular texture:

image
image
image

PR #399, opened 2023-11-15

kwcckw:


kwcckw:
Some examples of new granular texture:

image
image
image

PR #401, opened 2023-11-17

kwcckw:


kwcckw:
Some examples of new granular texture effect by stacking granules from larger size to smallest size:

image
image
image

PR #411, opened 2023-12-04
Added old paper color in PaperFactory.

kwcckw:


kwcckw:
Some examples of blending result using old paper color:

image
image
image
image

PR #412, opened 2023-12-05
Updated PaperFactory

kwcckw:
- Added preset "Blank" or "Old" color set in color texture.
- Added feature to select method in blending color into texture.
- Added feature to have user input color set in color texture.


kwcckw:
Examples of "Blank" color set:

image
image
image

Examples of user input color set:

image
image
image

PR #413, opened 2023-12-06
Added Moire effect augmentation.

### LINKED ISSUE #91: Create Moiré Pattern for Screen Pic Augmentation
It's common to use a camera phone to take a picture of a document displayed on a computer monitor. Often this is the lowest-friction means of making a record, which doesn't involve connecting a portable storage device or sending the file with email.

It might be good to have an augmentation that replicates the visual warping effect resulting from this.
Some example shots taken from this Issue submission page, taken with my phone (and JPEG compressed to fit GitHub's upload size requirements):
IMG_20210917_145957
IMG_20210917_150008

The warping effect, called [the Moiré pattern](https://en.wikipedia.org/wiki/Moir%C3%A9_pattern) is produced by interfering signals in the light produced by the screen pixel grid and the light received by the camera's sensor grid.

[This repository](https://github.com/gmelodie/silentMoire) contains code for *removing* the effect, which should be adaptable to add it back in.

kwcckw:


kwcckw:
Some examples:

image
image
image

PR #414, opened 2023-12-07
Added DoubleExposure augmentation.

kwcckw:


kwcckw:
Some examples:

image
image
image

PR #417, opened 2023-12-14
Added LensFlare.

kwcckw:


kwcckw:
The LensFlare augmentation is adapted from this link:

https://www.shadertoy.com/view/4sX3Rs

Some examples:
image
image
image

PR #431, opened 2024-04-02
Update noisetexturize.py

kukugpt:
add assert to make sure that there are no logical errors in the functionality


kwcckw:
The similar bug is resolved with this pull request:
https://github.com/sparkfish/augraphy/pull/432

Issue #435, opened 2024-07-17
Interop is broken after adding masks, keypoints and bounding boxes.


mezotaken:
MRE is in documentation, concerning Interop functionality [link](https://augraphy-doc.readthedocs.io/en/latest/doc/source/helper_and_utilities/interop.html)

```
Traceback (most recent call last):
  File "D:\Misc\Projects\plate_distortion\tttt.py", line 35, in <module>
    augmented_image = pipeline.augment(image)["output"]
                      ^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\Misc\Projects\plate_distortion\venv\Lib\site-packages\augraphy\base\augmentationpipeline.py", line 213, in augment
    data = self.augment_single_image(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\Misc\Projects\plate_distortion\venv\Lib\site-packages\augraphy\base\augmentationpipeline.py", line 419, in augment_single_image
    self.apply_phase(data, layer="post", phase=self.post_phase)
  File "D:\Misc\Projects\plate_distortion\venv\Lib\site-packages\augraphy\base\augmentationpipeline.py", line 740, in apply_phase
    result = augmentation(
             ^^^^^^^^^^^^^
TypeError: Interop.__call__() got an unexpected keyword argument 'mask'
```

I believe it could be fixed by adding those arguments to `Interop.__call__`

PR #436, opened 2024-07-17
Fix interop by adding missing arguments

mezotaken:
Potential fix for #435


kwcckw:
Thanks!

jboarman:
This is great to see contributions from the community!  It's been a small group of contributors so far.  Attention has grown recently, so hopefully we will see more community involvement like this!  Thanks again @mezotaken!

PR #441, opened 2024-09-10
Fix UnboundLocalError in Geometric

MarieRoald:
When I tried to use the `Geometric` augmentation with bounding boxes but without key-points, I got an `UnboundLocalError` due to the `remove_indices `-list not existing at line 212. This PR should hopefully fix that.


kwcckw:
Thanks!

PR #442, opened 2024-09-16
Update some augmentations to work with new version of numpy

MarieRoald:
I got errors when I tried to use the highlighter effect. It turned out that the std_range was integers but stored as floating point numbers. `random.randint` needs integers, so by explicitly casting to `int`, the code works again.


MarieRoald:
I encountered the same issue with `BadPhotoCopy`, so I casted those numpy-types to native types as well

kwcckw:
> I encountered the same issue with `BadPhotoCopy`, so I casted those numpy-types to native types as well

Thanks, could you show me the error again and the version of numpy that's causing the error?

PR #443, opened 2024-09-16
Minor improvements for geometric rotations

MarieRoald:
Two minor changes for the rotation:

1. Support for floating point rotation angles. I noticed it since I wanted <1° rotations (for very wide images).
2. Use the padding color as background for the rotation


kwcckw:
Thanks!

